PR #3, opened 2021-02-16
Remove access to Django and Wagtail admin login pages

rossmiller:
This was achieved through redirecting the relevant URLs.

Also included in this PR is the removal of the ability to set a user's password via Django admin and a session time out of 10 minutes.


SamDudley (approved):
Looks good to me :+1:

PR #5, opened 2021-02-17
WCM-229 Fix create_groups management command

SamDudley:


content/management/commands/create_groups.py
```diff
@@ -174,14 +166,31 @@ def handle(self, *args, **options):
...
-        #     viewers,
-        # ]:
-        #     GroupPagePermission.objects.get_or_create(
-        #         group=group,
-        #         page=home,
-        #         permission_type="edit",
-        #     )
+        # FIXME: remove
+        # news_permissions = [
+        #     # FIXME: do we need to add newscategory still?
```
SamDudley, line 171: Thoughts?

rossmiller: Yes, news categories should be added.

rossmiller (changes requested):
I've had a chat with the news team and they want to remove the concept of a news editor (they only need moderator).

rossmiller (approved):
LGTM!

PR #11, opened 2021-02-25
Hotfix/wcm 251 template tidy up

rossmiller:
- Tidy up CSS folder
- Restructure templates use of styles for header and top of column styles


SamDudley (approved):
Looks good to me :)

PR #35, opened 2021-03-17
WCM-312 Change news authors to be the first publisher

SamDudley:


content/models.py
```diff
@@ -88,6 +91,21 @@ def serve(self, request):
 
         return response
 
+    def get_first_publisher(self) -> Optional[UserModel]:
+        """Return the first publisher of the page.
```
oskusalerma, line 95: `"""Return the first publisher of the page or None."""`

oskusalerma (changes requested):

oskusalerma (approved):

PR #37, opened 2021-03-18
WCM-310 Add django simple history to non page models

SamDudley:


kevinetienne (approved):
LGTM. I've never used django simple history but I've checked the docs and it looks good. Can you see the history in the admin now?

davecharles (approved):
LGTM

PR #38, opened 2021-03-19
WCM-317 Upgrade wagtail and pillow

SamDudley:


kevinetienne (approved):
Looks good to me. I've checked pillow as it's a dependency from wagtaiil and it looks like they support it until v9 (https://github.com/wagtail/wagtail/blob/ee71bd3d07d1e70b335bc68e45f5a301613e568f/setup.py#L30).

davecharles (approved):
LGTM

PR #51, opened 2021-05-17
PFM-65 Add teams and roles to profile

SamDudley:


peoplefinder/views/role.py
```diff
@@ -0,0 +1,72 @@
+from django.contrib import messages
```
SamDudley, line 1: check over for potential refactor

peoplefinder/forms/role.py
```diff
@@ -0,0 +1,28 @@
...
+class RoleForm(forms.ModelForm):
+    class Meta:
+        model = TeamMember
+        fields = [
+            "person",
+            "team",
+            "job_title",
+            "head_of_team",
+        ]
+        widgets = {"person": forms.HiddenInput()}
```
kevinetienne, line 15: Is there any risk, if a user temper with the person pk in the form?

peoplefinder/views/role.py
```diff
@@ -0,0 +1,71 @@
...
+
+        return render(
+            request, "peoplefinder/components/role-edit.html", context=context
+        )
+
+    def delete(self, request, *args, **kwargs):
+        if self.role:
+            self.role.delete()
+
+        return HttpResponse("")
```
kevinetienne, line 64: Do you need a separate view to handle the deletion as the view is called `RoleEditView` and should you return a redirect? Do you have a constraint with the js? It looks like this view support both rendering html and responding to an ajax call for deletion.

peoplefinder/views/role.py
```diff
@@ -0,0 +1,71 @@
...
+from django.http import HttpResponse, JsonResponse
+from django.shortcuts import render
+
+from peoplefinder.forms.role import RoleForm
+from peoplefinder.models import Person, TeamMember
+from peoplefinder.services.team import TeamService
+from .base import PeoplefinderView
+
+
+class RoleEditView(PeoplefinderView):
```
kevinetienne, line 11: Who can edit role, do you need a permission check or see if the current user editing the role is allowed to change the roles?

kevinetienne (commented):
Added 3 comments which seems to revolve around the same topic (who can edit/delete roles).

peoplefinder/views/role.py
SamDudley: I do need to handle deletion and the empty response is for htmx to be able to remove the role from the DOM. I will add a comment explaining the empty response and will rename the view `RoleFormView`.

peoplefinder/forms/role.py
SamDudley: Yes there is, I'll add some validation to the post to check this.

peoplefinder/views/role.py
SamDudley: Added a basic check, the permissions in peoplefinder are interesting and will be implemented in detail at a later date.

kevinetienne (approved):
I didn't know about `UserPassesTestMixin` :+1:. Looks good to me but you might want to have someone as you mentioned to review the js.

PR #71, opened 2021-06-07
News pages do not have last viewed set

rossmiller:
Set last_viewed cookies for news pages and news home page.


SamDudley (approved):
LGTM!

PR #105, opened 2021-07-22
WCM-376 and WCM-377 Fix content issues

SamDudley:
- WCM-376 Fix page list for about us and tools
- WCM-377 Add ability to add child pages to about us pages

### DIFF
--- a/about_us/models.py
+++ b/about_us/models.py
@@ -4,8 +4,18 @@ from working_at_dit.models import PageWithTopics
 class AboutUs(PageWithTopics):
     is_creatable = True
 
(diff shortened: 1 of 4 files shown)


about_us/models.py
```diff
@@ -4,11 +4,31 @@
...
+        )
+
+        return context
+
 
 class AboutUsHome(PageWithTopics):
     is_creatable = False
 
     subpage_types = ["about_us.AboutUs"]
+
```
Luisella21, line 25: Line 25, 34 seems to be identical to line 10-19.
Is there a reason for the duplication?

Luisella21 (approved):
LGTM.

PR #106, opened 2021-07-22
PFM-101 Add profile completeness

SamDudley:


peoplefinder/views/profile.py
```diff
@@ -19,6 +19,7 @@ class ProfileDetailView(DetailView, PeoplefinderView):
     context_object_name = "profile"
     template_name = "peoplefinder/profile.html"
     pk_url_kwarg = "profile_pk"
+    queryset = Person.objects.with_profile_completion()
```
kevinetienne, line 22: as this is a class attr, would the query be cached? What happen if you update your profile without restarting the server, does the percentage update?

kevinetienne (approved):

peoplefinder/views/profile.py
SamDudley: As `with_profile_completion` only adds an annotation and is not a terminal ORM method, the queryset remains lazy so this is fine.

PR #115, opened 2021-08-17
PFM-157 Add legacy peoplefinder database connection

SamDudley:
- Tidy up readme
- Add legacy peoplefinder db connection
- fix db url parse
- add models and router
- fix router
- fix ABC
- add stub command


rossmiller (approved):
LGTM!

PR #169, opened 2022-01-31
Add pre-commit hooks

nahmed99:


rossmiller (approved):
LGTM!

config/settings/base.py
```diff
@@ -34,7 +34,7 @@
 # Feature flag to enable the new peoplefinder v2
 # TODO: Remove feature flag once we have migrated to v2 in prod
 PEOPLEFINDER_V2 = env.bool("PEOPLEFINDER_V2", False)
-PEOPLEFINDER_PROFILE_API_PRIVATE_KEY = b64decode(
+PEOPLEFINDER_PROFILE_API_PRIVATE_KEY = b64decode(  # /PS-IGNORE
```
SamDudley, line 37: Would this not be better in the exclude file? The comment looks out of place as the line has no resemblance to sensitive information. Same with the other comments. IMO the comments should be used when we have what looks like sensitive information, but we want to ignore it.

rossmiller: I think it would be dangerous to exclude the settings file because it's somewhere secrets/PII might be added by mistake.

rossmiller: @nahmed99 which check is this output from (NER or other)?

nahmed99: @rossmiller. I think this was from the other output (on terminal). The NER file doesn't contain this.

PR #178, opened 2022-02-15
Pfm 195 view log permissions

Luisella21:
New permission for View Team Log.
Tests for View Log and View Team Log


rossmiller (approved):
LGTM!

PR #196, opened 2022-03-21
Switch to OpenSearch

rossmiller:
This PR does not replace the ES packages we're using in the project as OpenSearch is compatible. The intention is that these will be replaced seperately.


SamDudley (commented):
LGTM!

SamDudley (approved):

PR #199, opened 2022-03-25
PFM 223 add person admin options

nahmed99:


peoplefinder/views/profile.py
```diff
@@ -65,7 +65,7 @@ def get_context_data(self, **kwargs: dict) -> dict:
             context["profile_audit_log"] = AuditLogService.get_audit_log(profile)
             context["profile_audit_log_excluded_keys"] = [
                 "user_id",
-                "manager_id",
+                "manager_id",  # /PS-IGNORE
```
SamDudley, line 68: I'm not a fan of the codebase and git diffs being littered with random comments to support the PII tool. Could this line go into the line based exclude file instead.

SamDudley (changes requested):

peoplefinder/views/profile.py
nahmed99: @SamDudley . Done.

SamDudley (approved):
LGTM!

PR #254, opened 2022-06-09
Add webhook for person post save updates to JML

rossmiller:


peoplefinder/models.py
```diff
@@ -185,6 +186,78 @@ class PersonManager(models.Manager):
...
+                                roles__head_of_team=True, then=Value(" (head of team)")
+                            ),
+                            default=Value(""),
+                        ),
+                    ),
+                    filter=Q(roles__isnull=False),
+                    distinct=True,
+                ),
+            )
+            .annotate(
```
jafacakes2011, line 211: Is there a reason for chaining all of these `.annotate()` methods?

From my understanding, all of these can be contained in one `.annotate()`.

jafacakes2011 (approved):
Just the one comment which is just a slight nitpick, other than that it LGTM!

peoplefinder/models.py
rossmiller: Thanks, I'll look into combining them.

PR #310, opened 2023-02-09
WCM-432 Convert the project over to Poetry

SamDudley:


Dockerfile
```diff
@@ -3,11 +3,15 @@ FROM python:3.9-buster
 ENV PYTHONUNBUFFERED=1
 ENV PYTHONDONTWRITEBYTECODE=1
 ENV PYTHONSTARTUP=.pythonrc.py
+ENV POETRY_VIRTUALENVS_CREATE=false
```
marcelkornblum, line 6: why not have poetry manage virtualenv? IMO makes it all much simpler...unless you want no venv at all, given we're in containers?

marcelkornblum (approved):

Dockerfile
SamDudley: The latter, given we are in a container there is no need for a venv in this case.

PR #316, opened 2023-02-15
WCM-43: add sanitize function for search queries

marcelkornblum:
People entering various non-URL-safe chars break the search function

https://sentry.ci.uktrade.digital/organizations/dit/issues/71022/?project=161

This PR removes those characters from the query silently, allowing for more likely matching. 

It also preserves quoted search phrases, but parses all their contents to remove unsafe chars. Phrases are only recognised if properly quoted inside matching pairs of quotes - either `'` or `"`. These quote marks are preservable because the legacy query building functionality treats them as a special case and processes them accordingly.

### DIFF
--- a/search/search.py
+++ b/search/search.py
@@ -2,6 +2,7 @@ from typing import Any, Optional
 
 from django.db.models import QuerySet
 from django.http import HttpRequest
+import re
 from wagtail.search.backends import get_search_backend
 from wagtail.search.utils import parse_query_string
 
@@ -51,3 +52,10 @@ def _person_query(request: HttpRequest) -> QuerySet:
 
 def _team_query(request: HttpRequest) -> QuerySet:
     return Team.objects.all()
+
+
+def sanitize_search_query(query: str) -> str:
+    if query is None:
+        return ""
+
+    return re.sub('[^a-zA-Z0-9-.~_\s]', "", query)
--- a/search/search.py
+++ b/search/search.py
@@ -1,4 +1,5 @@
 from typing import Any, Optional
+import unicodedata
 
 from django.db.models import QuerySet
 from django.http import HttpRequest
@@ -54,8 +55,12 @@ def _team_query(request: HttpRequest) -> QuerySet:
     return Team.objects.all()
 
 
-def sanitize_search_query(query: Optional[str]) -> str:
+def sanitize_search_query(query: Optional[str] = None) -> str:
     if query is None:
         return ""
 
-    return re.sub('[^a-zA-Z0-9-.~_\s]', "", query)
+    output = re.sub('[^a-zA-Z0-9-.~_\s]', "", query)
+    if not unicodedata.is_normalized("NFKD", output):
+        output = unicodedata.normalize("NFKD", output)
+
+    return output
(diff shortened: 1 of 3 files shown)


search/search.py
```diff
@@ -51,3 +52,10 @@ def _person_query(request: HttpRequest) -> QuerySet:
 
 def _team_query(request: HttpRequest) -> QuerySet:
     return Team.objects.all()
+
+
+def sanitize_search_query(query: str) -> str:
+    if query is None:
```
SamDudley, line 58: If `query` can be `None` then the type annotation should be `query: Optional[str]`.

search/search.py
```diff
@@ -51,3 +52,10 @@ def _person_query(request: HttpRequest) -> QuerySet:
...
+        return ""
+
+    return re.sub('[^a-zA-Z0-9-.~_\s]', "", query)
```
SamDudley, line 61: I think there will be issues with this if people want to search using accented characters?

SamDudley (changes requested):

search/search.py
marcelkornblum: good call. do we know what the ES tokenisation does - does it retain those or convert them to ascii?

or should i investigate?

search/search.py
```diff
@@ -51,3 +53,37 @@ def _person_query(request: HttpRequest) -> QuerySet:
...
+            quote_next_match = True  # opening quote found
+            continue
+        if match in valid_quotes and quote_next_match:
+            quote_next_match = False  # closing quote found
+            continue
+
+        # replace all url-unsafe chars
+        cleaned_match = re.sub(r"[^a-zA-Z0-9-.~_\s]", "", match)
+        if quote_next_match:
+            cleaned_match = f"'{cleaned_match}'"
```
SamDudley, line 82: Should the quotes be double or singles? Could the quotes match what the user put?

search/search.py
```diff
@@ -51,3 +53,37 @@ def _person_query(request: HttpRequest) -> QuerySet:
 
 def _team_query(request: HttpRequest) -> QuerySet:
     return Team.objects.all()
+
+
+def sanitize_search_query(query: Optional[str] = None) -> str:
```
SamDudley, line 58: I think this kind of change warrants a set of test cases to ensure the behaviour works as expected and does not regress if we need to modify it further.

SamDudley (changes requested):

search/search.py
marcelkornblum: AFAIK OpenSearch doesn't care; I thought I'd make them consistent to the search as it doesn't affect the user at all

search/search.py
marcelkornblum: tests added - and a good thing as well, I realised I was stripping rather than asciifolding 🤦

search/test/test_search.py
```diff
@@ -0,0 +1,99 @@
+from django.test import Client, TestCase
+
+from search.search import sanitize_search_query
+
+
+class TestSearchUtils(TestCase):
```
SamDudley, line 6: Suggestion: refactor using pytest's parametrize feature

https://docs.pytest.org/en/6.2.x/parametrize.html

example here https://github.com/uktrade/digital-workspace-v2/blob/9afb541512bd48d43a0c04d8bb3f21ae4e55b022/peoplefinder/test/services/test_audit_log.py#L94

SamDudley (approved):

PR #321, opened 2023-02-23
Add djlint config

marcelkornblum:
Added djlint to dev dependencies
add djlint config to pyproject.toml
ran full linting across the project, also fixing all errors


core/templates/includes/child-links.html
```diff
@@ -1,23 +1,33 @@
 <div class="ws-children-{{ num_cols }}-col">
+
     {% if children %}
+
         {% for child in children %}
             <div class="ws-col-child">
                 <h3 class="govuk-heading-s">
-                    <a {% if target_blank %}target="_blank" {% endif %}href="{{ child.url }}">{{ child.title }}</a>
+                    <a 
+                        {% if target_blank %}target="_blank"{% endif %}
```
jafacakes2011, line 9: Not something that needs changing in this commit, just something to discuss:
Personally, I prefer putting the non conditional attributes first.

I realise that this was just a reformat the existing code, just thought I'd put this out there. Maybe we could define some "good practices".

jafacakes2011 (approved):

marcelkornblum:
maybe we also want to remove the "preserve extra blank lines" thing and reformat again? @SamDudley @jafacakes2011

PR #327, opened 2023-03-01
Add reset DB command and improve makefile

jafacakes2011:


Makefile
```diff
@@ -74,46 +132,52 @@ webpack:
...
 pf-groups:
-	$(wagtail) python manage.py create_people_finder_groups
+	$(wagtail-exec) python manage.py create_people_finder_groups
 
 create_section_homepages:
-	$(wagtail) python manage.py create_section_homepages
+	$(wagtail-exec) python manage.py create_section_homepages
+
+reset-db:
+	docker-compose stop db
```
marcelkornblum, line 174:
```suggestion
	docker-compose kill db
```
not much in it but this might be slightly quicker?

marcelkornblum (approved):
I really like it. Does mean many commands will fail if you haven't got your containers already running - does anyone mind that?

jafacakes2011:
Yea, I guess it depends on the philosophy you want to take with the makefile commands.

Just took a stab at adding a check to see if the container is running, so we can call with a preference for exec if available, `makemigrations` would change from:

```
makemigrations:
    $(wagtail-run) python manage.py makemigrations
    $(chown) */migrations/*
```

to

```
makemigrations:
    @if [ -z `docker ps -q --no-trunc | grep $(docker-compose ps -q wagtail)` ]; then \
        $(wagtail-run) python manage.py makemigrations; \
    else \
        $(wagtail-exec) python manage.py makemigrations; \
    fi
    $(chown) */migrations/*
```

marcelkornblum:
> if there's a way we could make it so that we can just call it like this: `$(wagtail) python manage.py makemigrations` and it contains that condition then i think this might be a good option. Open to thoughts.

if you can make that work it'd be awesome, and we'd use it everywhere. I had a fleeting thought of doing it but got scared away from it by the whole "getting makefile to do more than trivial tasks" thing, which is not my expert area :)

marcelkornblum:
incorporated into #358

marcelkornblum:
closing in favour of #358 since the "use running container if poss" is sorted there and all the other changes from here are also brought across

PR #346, opened 2023-03-28
Add more e2e tests

jafacakes2011:


e2e_tests/test_wagtail_admin.py
```diff
@@ -0,0 +1,82 @@
...
+from .utils import login
+
+
+@pytest.mark.e2e
+def test_add_news_page(superuser, page: Page):
+    news_home = NewsHome.objects.first()
+    NewsPageFactory.create_batch(5, parent=news_home)
+
+    login(page, superuser)
+    page.goto("/admin")
```
marcelkornblum, line 16: we should probably drop the smoke test that feels overlapping with this and consolidate i reckon

marcelkornblum (approved):

PR #350, opened 2023-04-18
headings and result types split out for extra uses

marcelkornblum:
sometimes we need the result type for things like "featured guidance pages" or "there are no tools matching your search" UI elements even if we're not showing headings


jafacakes2011 (approved):
LGTM! 😀

PR #351, opened 2023-04-18
two-column layout for page type results

marcelkornblum:
there may be a more extensible way to do this but i think seems fine right now ?


jafacakes2011 (approved):
LGTM!

PR #365, opened 2023-05-05
DWPF-501 Fix deleting a profile without a user

SamDudley:


marcelkornblum (approved):
nice!

do you know how we ended up with profiles without users in the first place?

PR #369, opened 2023-05-30
containers not relying on py-node

marcelkornblum:
M1 macs struggle with the non-multi-arch nature of py-node; until that's sorted regressing to local build but using py-node dockerfile commands

also minor patch updates for python deps, and linting fixes that got noticed after those updates


jafacakes2011 (approved):
LGTM!

PR #372, opened 2023-06-06
Search improvements: building blocks and content indexing

marcelkornblum:
Note that user-facing functionality should not be affected (or only very minimally) by this PR, except that the content editor page admin interface has the "excerpt" field at the top, and with a large text area, and that that field is what users always see on page search results.

* Create and [document](../blob/main/docs/search.md) overall strategy for search indexing
* Build new extended_search component (see below)
* Rework document indexing to capture each important field in all indexed models (it's possible but unlikely that this will affect user-facing functionality)
* _Don't affect existing search query functionality_
* Add new permissions to edit search boosting weights
* Add basic new "explore" page so people with those permissions can see the new search
* Add "excerpt" field to Content Page, migrate existing PageWithTopics and NewsPage excerpt field contents over
* Ensure excerpt field is what shows in search results summary 

Extended search component extends Wagtail and (in this PR):
* breaks functionality into smaller pieces to make overriding easier
* adds a new OnlyFields wagtail SearchQuery to limit the extent of nested SearchQuery pieces
* Creates a Mixin and uses it for all Wagtail index.SearchField and similar, allowing for a new parameter so the name doesn't have to match the model field name
* Creates new IndexedField class to capture the relationship between mode- and index-fields (and allow e.g. one-to-many)
* Creates "managers" which 
  * allow search field definition in a custom model, using new IndexedField
  * Generate wagtail-compliant search_fields attribute using mixed-in classes
  * auto-generate wagtail SearchQuery DSL based on IndexedField mappings and search input

Things that are obviously still @TODO (and for another PR):
* Getting the settings working peroperly
* Getting the query generation working properly, especially for relatedFields
* sorting out proximity boosting
* possible refactoring o the IndexManager class inheritance to be more sensible
* all content-specific boostings
* other things as well!!


marcelkornblum:
@jafacakes2011 @SamDudley  @rossmiller want a session for me to take you through this monster? Would love to get it merged as soon as poss

src/extended_search/index.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+class Indexed(index.Indexed):
+    @classmethod
+    def _check_search_fields(cls, **kwargs):
+        errors = []
+        for field in cls.get_search_fields():
+            message = "{model}.search_fields contains non-existent field '{name}'"
+            if not cls._has_field(field.field_name) and (
+                "kwargs" in field.__dict__
```
marcelkornblum, line 19: use hasattr instead

src/networks/migrations/0003_remove_network_excerpt.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("networks", "0002_network_excerpt"),
+        ("content", "0018_remove_contentpage_body_no_html_and_more"),
+    ]
+
+    operations = [
+        migrations.RunSQL(
```
marcelkornblum, line 13: investigate doing it in python using wagtail builting methods for finding parents etc

src/peoplefinder/models.py
```diff
@@ -511,6 +642,30 @@ def all_languages(self) -> str:
...
+        abbrs = teams.values_list("team__abbreviation", flat=True)
+        abbrs_str = " ".join(list([a or "" for a in abbrs]))
+        return f"{names_str} {abbrs_str}"
+
+    @property
+    def search_titles(self):
+        return ", ".join(self.roles.all().values_list("job_title", flat=True))
+
+    @property
+    def search_email(self):
```
marcelkornblum, line 666: delete this

src/search/search.py
```diff
@@ -25,16 +38,13 @@ class PagesSearchVector(SearchVector):
...
-            .not_pinned(query)
-            .search(query, *args, operator="and", **kwargs)
-        )
-
     def pinned(self, query):
         return self.get_queryset().pinned(query)
 
+    def search(self, query, *args, **kwargs):
+        queryset = self.get_queryset().not_pinned(query)
+        return self._wagtail_search(queryset, query, *args, **kwargs)
```
marcelkornblum, line 46: add operator and back in

src/search/search.py
```diff
@@ -68,7 +78,51 @@ def get_queryset(self):
 
         return people
 
+    def search(self, query, *args, **kwargs):
+        queryset = self.get_queryset()
```
marcelkornblum, line 82: add not_pinned
and and operator

src/working_at_dit/migrations/0006_remove_pagewithtopics_excerpt.py
```diff
@@ -0,0 +1,21 @@
...
+    dependencies = [
+        ("working_at_dit", "0005_rebrand_to_dbt"),
+        ("content", "0018_remove_contentpage_body_no_html_and_more"),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            sql="UPDATE content_contentpage AS cp SET search_excerpt = (SELECT excerpt FROM working_at_dit_pagewithtopics AS wadpwt WHERE wadpwt.contentpage_ptr_id = cp.basepage_ptr_id);",
+            reverse_sql="UPDATE working_at_dit_pagewithtopics AS wadpwt SET excerpt = (SELECT search_excerpt FROM content_contentpage AS cp WHERE wadpwt.contentpage_ptr_id = cp.basepage_ptr_id);",
+        ),
```
marcelkornblum, line 16: see other migration - python

marcelkornblum:
@jafacakes2011 @SamDudley would you mind taking a look now please? I've made the changes we discussed on Friday...

src/extended_search/index.py
```diff
@@ -66,28 +68,19 @@ def get_definition_model(self, cls):
...
         Returns the value from the model's field if it wasnt found because of a naming discrepancy
         """
-        value = super().get_value(obj)
-        if (
-            value is None
-            and "kwargs" in self.__dict__
-            and "model_field_name" in self.kwargs
-        ):
-            value = getattr(obj, self.kwargs["model_field_name"], None)
+        if not (value := super().get_value(obj)):
```
jafacakes2011: I wonder if this reads better like so:
```
if value := super().get_value(obj):
    return value
return getattr(obj, self._get_model_field_name(), None)
```
Not a dealbreaker though

jafacakes2011 (approved):
Just one small comment

PR #376, opened 2023-06-19
New models for irap

Luisella21:
- First pass at creating the new models required for importing IRAP tools from Data workspace.
- Really small changes to templates, to allow viewing the Tools data.
- Primitive admin interface, to be used for testing.


src/tools/models.py
```diff
@@ -6,20 +8,95 @@
...
+        REVIEWED = "reviewed", "Reviewed"
+
+    after_import_status = models.CharField(
+        max_length=9, choices=AfterImportStatus.choices, default=AfterImportStatus.NEW
+    )
+    review_date = models.DateTimeField(null=True, blank=True)
+
+    imported = models.BooleanField(default=True)
+
+    created_date = models.DateTimeField(auto_now_add=True)
```
jafacakes2011, line 64: It might be handy to have a field that tracks when the object was last updated

src/tools/models.py
```diff
@@ -6,20 +8,95 @@
...
+
+    # TODO Move it to a util file
+    readonly_widget = forms.TextInput(attrs={"disabled": "true"})
 
     parent_page_types = ["tools.ToolsHome"]
     subpage_types = []  # Should not be able to create children
 
     content_panels = PageWithTopics.content_panels + [
         FieldPanel("redirect_url"),
+        FieldPanel("irap_tool", widget=readonly_widget),
```
jafacakes2011, line 99: Instead of defining `readonly_widget` you could move it to a util file as you mentioned, or you could just define it inline:

```python
FieldPanel(
    "irap_tool",
    widget=forms.TextInput(attrs={"disabled": "true"}),
),
```

src/tools/models.py
```diff
@@ -6,20 +8,95 @@
...
+        blank=True,
+    )
+
+    # TODO Add new fields that will be created later:
+    #  description and condition of use
+    class Meta:
+        abstract = True
+
+
+class IrapToolDataImport(IrapToolDataAbstract):
```
jafacakes2011, line 35: I'm not too sure about the 2 step import with 2 models. But that might be because I'm slightly out of the loop on recent conversations.

jafacakes2011 (commented):
Just some initial comments 😄

src/tools/models.py
Luisella21: I like it!

src/tools/models.py
Luisella21: Added a comment explaining why.

jafacakes2011 (approved):
Looking good!

PR #380, opened 2023-07-04
People Finder Profile updates (DWPF-587 DWPF-588 DWPF-589)

jafacakes2011:


assets/stylesheets/components/_peoplefinder.scss
```diff
@@ -380,3 +380,10 @@
     width: 50%;
   }
 }
+
+.missing-profile-completion-fields {
+  // Add a comma after each field except the last, unless there is only one.
+  > .missing-profile-completion-field:not(:last-child):not(:only-child):after {
```
marcelkornblum, line 386: 👍 for approach
😂 for class name

jafacakes2011: Yea, I want a better name, waiting for a decision to be made on what we tag them. (currently "recommended", but this is going to change)

src/peoplefinder/templates/peoplefinder/profile-edit.html
```diff
@@ -82,7 +122,7 @@ <h2 class="govuk-heading-m">Contact details</h2>
         </fieldset>
 
         <fieldset>
-            <h2 class="govuk-heading-m">Location and working patterns</h2>
+            <h2 id="cams-test" class="govuk-heading-m">Location and working patterns</h2>
```
marcelkornblum, line 125: needs a proper ID?

marcelkornblum (commented):
just a couple of comments (other than the inline ones):
- I'm surprised we're not reordering the edit page more; is this really it?
- something else that i cant remember but if i  do i'll update here :(

src/peoplefinder/templates/peoplefinder/profile-edit.html
jafacakes2011: This is debug stuff that I need to remove, good catch!

src/peoplefinder/utils.py
```diff
@@ -0,0 +1,42 @@
...
+    "roles",  # Related objects
+]
+
+
+def get_profile_completion(person: "Person") -> int:
+    complete_fields = 0
+    field_statuses = profile_completion_field_statuses(person)
+    for field_status in field_statuses:
+        if field_statuses[field_status]:
+            complete_fields += 1
```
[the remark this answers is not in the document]
jafacakes2011: TBH, I think we should remove required fields like first name, last name and email from the calculation. Feels odd that their included if they are required.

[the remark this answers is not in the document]
marcelkornblum: true, and agreed (but I still think a weighting would be useful :P )

SamDudley (approved):

PR #381, opened 2023-07-04
Update action to be triggered by dependabot PRs

marcelkornblum:
#373 appears not to result in an action being triggered - perhaps dependabot doesn't `push` code?


jafacakes2011 (approved):
Interesting find. Keen to see this working!

marcelkornblum:
> Interesting find. Keen to see this working!

we won't really know until the next dependabot event :/

PR #382, opened 2023-07-05
Feature/tools 10 process imported

Luisella21:
Implement the logic used when importing a new set of data for IRAP tools.
It assumes that the new data has already been validated, and it exists in an intermediate table.


src/tools/import_irap_tool.py
```diff
@@ -0,0 +1,113 @@
...
+            return False
+    return True
+
+
+def process_import():
+    """To be called after the irap data has been imported
+    successfully from Data Workspace.
+    There is no validation on the new records, as any validation will
+    happen when the data is imported from DW
+    """
```
marcelkornblum, line 37: I'm left confused by this docstring I'm afraid. Has the data already been validated at this point or is it still to be validated? "DW" is also confusing in  this context :)

src/tools/import_irap_tool.py
```diff
@@ -0,0 +1,113 @@
+from tools.models import IrapToolData, IrapToolDataImport, Tool
```
marcelkornblum, line 1: I thought the plan was to build this in a way that would easily extensible if/when we want to import from other sources in the future...

Should we be calling this module `import.py` and making all the methods a bit less specific to IRAP, unless they are specifically and only for IRAP data?

src/tools/import_irap_tool.py
```diff
@@ -0,0 +1,113 @@
...
+    # are reviewed
+    # The format may not be the best, but it can be reviewed when
+    # it  has to be displayed
+
+    new_fields = new._meta.get_fields()
+    changed = False
+    previous_values = {}
+    for field in new_fields:
+        new_value = getattr(new, field.name)
+        old_value = getattr(old, field.name)
```
marcelkornblum, line 16: do we need to catch times when the field list and/or names might have changed? i.e. `old` might not have a field that exists on `new`, or the other way around.

src/tools/import_irap_tool.py
```diff
@@ -0,0 +1,113 @@
...
+        if old_value != new_value:
+            changed = True
+            setattr(old, field.name, new_value)
+        previous_values[field.name] = old_value
+
+    return changed, previous_values
+
+
+def are_field_identical(import_obj: IrapToolDataImport, old_values: dict) -> bool:
```
marcelkornblum, line 25:
```suggestion
def are_fields_identical(import_obj: IrapToolDataImport, old_values: dict) -> bool:
```

src/tools/import_irap_tool.py
```diff
@@ -0,0 +1,113 @@
...
+def process_import():
+    """To be called after the irap data has been imported
+    successfully from Data Workspace.
+    There is no validation on the new records, as any validation will
+    happen when the data is imported from DW
+    """
+    IrapToolData.objects.update(imported=False)
+
+    imported_iraps = IrapToolDataImport.objects.all()
+    for imported_irap in imported_iraps:
```
marcelkornblum, line 41: It would be great IMO to have all the field names in a configurable mapping object, so that it's easy to update which fields go to which if the column names in the data source change from time to time

marcelkornblum (commented):
Looking good! I didn't review the tests though - do you need me to?

CamLamb:
Marking this as a draft as the work hasn't been actively worked on in a while

PR #390, opened 2023-07-17
Fix: make migration into best effort

marcelkornblum:
some pages have been locked, and sometimes by nonexistent users, which is causing issues with the new migration


jafacakes2011 (approved):
LGTM!

PR #391, opened 2023-07-17
Search improvements: overridable settings

marcelkornblum:
A search settings module that allows frankly ridiculous amounts of override (this is merited because there are MANY settings and they pertain to different types of thing since it's a complex area).

Defaults are set in-p[lace in the settings module.

Any or all of the defaults nested dict can be overridden in the Django settings. Sparse dicts (i.e. only specifying the overrides) are fine; we use ChainMap but we've extended it to be infinitely nestable.

After that, any individual setting can be overridden by and ENV var, which can be overridden by an IndexedField's boost setting (only for field boost parts, of course), which can then be overridden by a DB setting.  Individual settings like this are definable with a mapping of `__` joining levels of the nested dict's keys - for example 
```py
"SEARCH_EXTENDED": {
    "boost_parts":{
        "search_phrase": 2.5
    }
}
```
becomes 
```py
SEARCH_EXTENDED__boost_parts__search_phrase=2.5
```

*NB. Disabling all related field indexing temporarily. For critical fields we have a work around with a flattened list field instead*
Related Fields causing issues with boost value retrieval, and more generally. Hope this is more sovleable as part of cleanup once query generation makes more sense.


src/search/views.py
```diff
@@ -41,10 +41,13 @@ def explore(request: HttpRequest) -> HttpResponse:
...
 
     boost_vars = [
-        {"name": f"SEARCH_BOOST_{k}", "value": v}
-        for k, v in extended_search_settings.BOOST_VARIABLES.items()
+        {"name": k, "value": extended_search_settings[k]}
+        for k in extended_search_settings.all_keys
+        if "boost_parts" in k
     ]
 
+    # boost_vars = extended_search_settings["boost_parts"]["fields"]
```
jafacakes2011, line 49: Commented out code

src/extended_search/signals.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+def update_searchsetting_queryset(sender, **kwargs):
+    extended_search_settings.initialise_db_dict()
+
+
+post_save.connect(update_searchsetting_queryset, sender=Setting)
+post_delete.connect(
+    update_searchsetting_queryset, sender=Setting
+)  # this seems to fire too early? not having an effect...
```
jafacakes2011: Is this still the case or can this comment be removed?

src/extended_search/admin.py
```diff
@@ -0,0 +1,25 @@
+from django import forms
+from django.contrib import admin
+
+from core.admin import admin_site
+from extended_search.models import Setting
+from extended_search.settings import extended_search_settings
+
+
+class SettingAdminForm(forms.ModelForm):
+    # doesn't contain fields at the time this is run...
```
jafacakes2011, line 10: You can move the choices into the `__init__` method setting `self.fields["key"].choices`
This will load the choices when the form is initialised instead of when the server is started.

src/extended_search/managers/query_builder.py
```diff
@@ -1,10 +1,11 @@
 import logging
+from django.contrib.contenttypes.models import ContentType
 from django.db import models
 from wagtail.search.query import Boost, Phrase, PlainText
 
 from search.utils import split_query
 from extended_search.backends.query import OnlyFields
-from extended_search.settings import DEFAULTS as settings
+from extended_search.settings import extended_search_settings as settings
```
jafacakes2011, line 8: Should we name this something other than `settings` as this is usually the Django settings var name, so might be confusing to someone skimming the code.
I'm not opposed to referencing `extended_search_settings` directly or perhaps something shorter like `xsearch_settings`

jafacakes2011 (approved):
LGTM!

PR #397, opened 2023-07-25
Add team select search

jafacakes2011:
Adds the team selector updates:
- Matches new designs
- Ordered by name (with options that have children going to the top)


assets/stylesheets/components/_peoplefinder.scss
```diff
@@ -287,54 +287,60 @@
   color: govuk-colour("red");
 }
 
-// Team select web component styles
-.team-select__teams {
-  margin-bottom: 20px;
-  border: 2px solid black;
+// Team select
+.team-select {
+  font-family: "GDS Transport", arial, sans-serif;
```
marcelkornblum, line 292: why do we need to add a new font at this level?

assets/stylesheets/components/_peoplefinder.scss
```diff
@@ -287,54 +287,60 @@
...
 
-  ul {
-    list-style: none;
-    margin: 0;
-    padding: 0;
+  &__teams {
+    max-height: 800px;
```
marcelkornblum, line 295: I'm not sure this is wrong in the context of the rest of the CSS, but it grinds my gears a bit having arbitrary numbers specified in px. Would it not make sense as a relation to either the viewport or the line height?

assets/stylesheets/components/_peoplefinder.scss
```diff
@@ -287,54 +287,60 @@
...
+    overflow-y: auto;
+    margin-bottom: 20px;
```
marcelkornblum, line 297: same for these measurements - probably not worth updating but don't we have a set of standardised margins etc accessible via scss var?

package.json
```diff
@@ -1,35 +1,38 @@
...
+        "last 1 firefox version",
+        "last 1 safari version",
+        "IE 11"
+    ],
     "engines": {
         "npm": ">8.0.0",
         "node": ">18.0.0"
+    },
+    "prettier": {
+        "printWidth": 100
```
marcelkornblum, line 36: not sure i agree with just adding in linting overrides ...

marcelkornblum (approved):
I didn't really review the JS because it was complex to start with, and the linting made it super hard to see what to review... otherwise it's OK though there are a couple of comments :)

assets/stylesheets/components/_peoplefinder.scss
jafacakes2011: Looks like we don't, I assume initially it was needed as the markup didn't have any gov classes

PR #405, opened 2023-07-26
Search improvements: query generation

marcelkornblum:
Tweaks the query generation implemented in #372 to implement Fuzzy, and to work better in general :) 

Doesn't address proximity boosting or filtering (those are upcoming enhancements)


src/content/models.py
```diff
@@ -82,11 +82,36 @@ def get_first_publisher(self) -> Optional[UserModel]:
...
+
+        q = Q()
+        for restriction in (
+            PageViewRestriction.objects.filter(restriction_type__in=types)
+            .select_related("page")
+            .all()
+        ):
+            q |= self.descendant_of_q(restriction.page, inclusive=True)
+
+        # do not match any page if no private section exists.
```
jafacakes2011, line 104: This comment needs updating as the function isn't just referring to private/public restrictions

jafacakes2011 (approved):
LGTM, just one comment

PR #406, opened 2023-07-26
Fix E2E tests

jafacakes2011:


jafacakes2011:
Merging even though there is one failing e2e test. I have tested the scenario locally and it works fine.

PR #416, opened 2023-08-01
Search improvement tweaks

marcelkornblum:
small tweaks to the way the new search and specifically the explore page work in terms of the new search, based on feedback from the hotfix env


CamLamb (approved):
LGTM!

PR #430, opened 2023-08-21
switch settings to en-gb and Europe/London TZ

marcelkornblum:
https://uktrade.atlassian.net/browse/DWPF-161

Not 100% convinced this is a great idea, though it's definitely nicer user experience for e.g. article published date. Any watch-outs you can think of, perhaps re: infra, logging, etc?


CamLamb:
It might be worth double checking, but I think `TIME_ZONE` might be used when storing dates/datetimes.
If it is, then you might want to either:
- write a migration (could be a task and a half!)
- See if there is a way to set the default `current time zone` https://docs.djangoproject.com/en/4.2/topics/i18n/timezones/#default-time-zone-and-current-time-zone

CamLamb (approved):
LGTM, nice n simple!

PR #432, opened 2023-08-29
Get RelatedFields working properly

marcelkornblum:
- [x] Fix issue so RelatedFields allow indexing and searching to happen without error
- [x] Add functionality so RelatedFields are included in QueryBuilder and properly queried
- [x] Add all model RelatedFields back in to all models
- [x] Remove all model properties used as workarounds for this


docker-compose.yml
CamLamb, line 1: Looks like this is just a reorder of the compose file, I quite liked redis/celery/celery-beat being alongside each other.

src/extended_search/backends/backend.py
```diff
@@ -35,11 +35,22 @@ def _remap_fields(self, fields):
         searchable_fields = {
             f.field_name: f for f in self.queryset.model.get_searchable_search_fields()
         }
+        if hasattr(self.queryset.model, "get_related_search_fields"):
+            for f in self.queryset.model.get_related_search_fields():
+                searchable_fields[f.field_name] = f
```
CamLamb, line 40: There's some trusting here that the f.field_name doesn't already exist in `searchable_fields`. A sensible assumption, but raising an exception might be nice?

CamLamb (approved):
LGTM! Just some lil comments

src/extended_search/backends/backend.py
marcelkornblum: I actually thought it might just override but that that wouldn't matter - it's a more naive way of writing the same code as in the previous lines, but for related fields rather than searchable ones

docker-compose.yml
marcelkornblum: celery and celery beat i understand; why redis? i like that all the dependencies and non-custom stuff is together and not split by wagtail :)

src/extended_search/backends/backend.py
marcelkornblum: in the end I reworked this to be more in line with the original implementation instead :)

PR #433, opened 2023-08-29
DWPF-704 update dw usage of feedback package and implement custom notifications

nicopicchio:


src/feedback/utils.py
```diff
@@ -0,0 +1,35 @@
...
+from notifications_python_client.notifications import NotificationsAPIClient
+
+
+def is_feedback_received_past_24hrs():
+    current_datetime = timezone.now()
+    time_difference = timedelta(days=1)
+    is_feedback_received = BaseFeedback.objects.filter(
+        submitted_at__gte=current_datetime - time_difference
+    ).exists()
+    return is_feedback_received
```
CamLamb, line 14: This method could be simplified a lil bit:

```python
def is_feedback_received_past_24hrs():
    return BaseFeedback.objects.filter(
        submitted_at__gte=timezone.now() - timedelta(days=1)
    ).exists()
```

src/feedback/utils.py
```diff
@@ -0,0 +1,35 @@
+from django.utils import timezone
+from django_feedback_govuk.models import BaseFeedback
+from datetime import timedelta
+from django.conf import settings
+from notifications_python_client.notifications import NotificationsAPIClient
+
+
+def is_feedback_received_past_24hrs():
```
CamLamb, line 8: I'm trying to think of a good name for this method, it's always the hardest part 😄 

I'm thinking something like `has_received_feedback_within_24hrs`?

src/feedback/utils.py
```diff
@@ -0,0 +1,35 @@
...
+    )
+    email_recipients = settings.FEEDBACK_NOTIFICATION_EMAIL_RECIPIENTS
+
+    if len(email_recipients) < 1:
+        raise Exception("No feedback recipients configured")
+
+    for email_recipient in email_recipients:
+        message = notification_client.send_email_notification(
+            email_address=email_recipient,
+            template_id=settings.FEEDBACK_NOTIFICATION_EMAIL_TEMPLATE_ID,
```
CamLamb, line 29: Let's look at the template on Notify and make changes if needed

src/feedback/utils.py
CamLamb, line 8: We should write some tests for this, let's hop on a call when we have some spare time and we can go over testing in Django/Python

CamLamb (commented):
When we merge PRs, their title become the commit message on the main branch.
Since this is the case, I like to make sure the ticket number is in the format that will be picked up by Jira:

`Dwpf 704` > `DWPF-704`

src/feedback/utils.py
nicopicchio: yeah I was debating whether to return directly or not. Changed it to the more succint version.

src/feedback/utils.py
nicopicchio: how about `was_new_feedback_received(number_of_days)` and pass a parameter for the `timedelta(days=number_of_days)`
e.g. `was_new_feedback_received(1)`

nicopicchio: method renamed to `feedback_received_within()` and default value set at 1.

nicopicchio: tests for feedbacks on db created. tests for email notifications still outstanding.

nicopicchio: tests for email notifications created

src/feedback/test/test_utils.py
```diff
@@ -0,0 +1,110 @@
...
+from unittest.mock import call
+from django.test import override_settings
+from django_feedback_govuk.models import BaseFeedback
+
+from feedback.utils import feedback_received_within, send_feedback_notification
+
+
+def test_no_feedback_submitted(db):
+    # feedback_received_within() returns False if no feedback was submitted.
+    assert not feedback_received_within()
```
marcelkornblum, line 11: I'd ideally merge this and the next test, as putting them together better explains what the feedback_received_within() method does and means less comments are needed

src/feedback/test/test_utils.py
```diff
@@ -0,0 +1,110 @@
...
+    # feedback_received_within() returns True if feedback was submitted within the last 24hrs
+    BaseFeedback.objects.create()
+    assert feedback_received_within()
+
+
+def test_feedback_submitted_over_24hrs_ago(db, freezer):
+    # feedback_received_within() returns False if feedback was submitted over 24hrs ago
+    freezer.move_to("2023-09-01")
+    BaseFeedback.objects.create()
+
```
marcelkornblum, line 24: add an assert here to be able to verify the change of status of the method return value

src/feedback/test/test_utils.py
```diff
@@ -0,0 +1,110 @@
...
+    freezer.move_to("2023-09-03")
+    assert not feedback_received_within()
+
+
```
marcelkornblum, line 28: the following ones are all fine, but very repetitive - if you want you could consider parameterising them to make it a bit more concise and (IMO) easier to see what's happening between the tests. There's examples of this in DW somewhere if you search around.

marcelkornblum (approved):
Looks good! There are a few things in the tests I'd ideally improve but no blockers

src/feedback/test/test_utils.py
nicopicchio: merged the two tests

src/feedback/test/test_utils.py
nicopicchio: assert added

PR #435, opened 2023-09-01
POC: Organogram

marcelkornblum:
Not for production use as it stands, just for demos. May want to feature-hide the organogram and keep the rest?

* Includes an organogram view/tpl for visualising the manager/report structure added to the profile view page
* updates the team page to simplify viewing members on that page 
* Updates the way Node is installed on the playwright container (old way is deprecated)


CamLamb:
I've just spent a quick moment bringing this branch up to date with main in case we want to do something with it in the future:
Screenshot 2025-04-16 at 15 41 39

PR #441, opened 2023-09-05
Profile complete banner on view page

CamLamb:


marcelkornblum (approved):
I'm just going to assume there aren't any legacy `first_name` fields in any templates, and that any properties using it (is there a `full_name`?) are updated... :)

PR #442, opened 2023-09-06
DWPF-746 first_name migration

CamLamb:
We don't need to ask Staff SSO for the the First Name as the User is created with the first_name from Staff SSO and it's unlikely that it has changed.

### DIFF
--- a/src/peoplefinder/migrations/0109_person_preferred_first_name.py
+++ b/src/peoplefinder/migrations/0109_person_preferred_first_name.py
@@ -4,6 +4,9 @@ from django.db import migrations, models
 
 
 def populate_preferred_first_name(apps, schema_editor):
+    """
+    Update all Profile models preferred_first_name with the profile.first_name
+    """
     Person = apps.get_model("peoplefinder", "Person")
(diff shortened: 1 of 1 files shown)


src/peoplefinder/migrations/0109_person_preferred_first_name.py
```diff
@@ -4,10 +4,24 @@
...
 
 def populate_preferred_first_name(apps, schema_editor):
+    """
+    Update all Profile models preferred_first_name with the profile.first_name
+    """
     Person = apps.get_model("peoplefinder", "Person")
     Person.objects.update(preferred_first_name=models.F("first_name"))
 
 
+def update_profile_first_name(apps, schema_editor):
```
CamLamb, line 14: This migration might take a while. We couldn't do this with     `Person.objects.update(first_name=models.F("user__first_name"))`
We aren't allowed to do Joined field references in migrations:
`django.core.exceptions.FieldError: Joined field references are not permitted in this query`

I'm going to rewrite this update as SQL

SamDudley (approved):

PR #444, opened 2023-09-11
DWPF-746 Fix e2e tests

CamLamb:
Test changes since we use the Preferred name now instead of the "first_name"


marcelkornblum (approved):
I can see they've just run successfully, but WTAF? How did this changeset make it work?

PR #446, opened 2023-09-11
DWPF-735 Add celery beat

CamLamb:


src/config/celery.py
```diff
@@ -11,11 +11,12 @@
 
 
 celery_app.conf.beat_schedule = {
-    # Nightly ingest tasks.
-    "ingest-uk-staff-locations": {
-        "task": "core.tasks.ingest_uk_staff_locations",
-        "schedule": crontab(minute="0", hour="3"),
-    },
+    # Staff location ingest task
+    # TODO: Uncomment this when the Staff Location DB is ready.
```
CamLamb, line 15: This is temporarily commented out, while we wait for the data team to set up the DB so we have somewhere to ingest from

marcelkornblum (approved):
LGTM

PR #448, opened 2023-09-11
DWPF-734 Adapt MoJ side navigation to mobile screen

nicopicchio:


src/peoplefinder/templates/peoplefinder/profile-edit.html
```diff
@@ -41,9 +40,10 @@ <h1 class="govuk-heading-l">{{ page_title }}</h1>
                     {% endfor %}
                 </ul>
             </nav>
-
+
```
CamLamb, line 43: What's with all the spaces?

CamLamb (approved):
PR looks good, I'll get this deployed to dev so we can review the style changes

PR #457, opened 2023-09-19
DWPF-773 Adds days in the office field

nicopicchio:


src/peoplefinder/models.py
```diff
@@ -791,6 +798,11 @@ def get_workdays_display(self) -> str:
         # "Monday, Tuesday, Wednesday, ..."
         return ", ".join(map(str, workdays))
 
+    def get_usual_office_days_display(self) -> str:
```
CamLamb, line 801: This isn't needed, if all you are doing is printing the value of the field, instead of defining `get_usual_office_days_display` in the Tuple, you can just use the field name.

I built the Tuple so that the first key is either an attribute or callable on the model, this way you can just return the field value, or if you need to alter the value for presentation purposes, you can define a method.

In this scenario it looks like you are just using the field value.

nicopicchio: ok, sorted.

PR #458, opened 2023-09-19
Update documentation to use mkdocs

CamLamb:


marcelkornblum (commented):
should we limit the action to only run when appropriate - like only when the pr includes changes to /docs ?

marcelkornblum (approved):

PR #461, opened 2023-09-21
DBTP-457 Use dbt-copilot-python to set ALLOWED_HOSTS & DATABASES in DBT PaaS

WillGibson:
What it says in the title, plus it should remain compatible with GOV.UK PaaS.


CamLamb (approved):
LGTM.
Note: that this will deploy to staging on merge to main

PR #464, opened 2023-09-22
DBTP-457 Add DBT PaaS CodeBuild configuration

WillGibson:


user-post.sh
CamLamb, line 1: The name of this script doesn't seem to match what it is doing.
Also, should this exists somewhere other than the project root?

CamLamb:
I'm not too happy with the settings situation, I think we might want to reconsider the approach here or at lease use some better names, having 2 base files doesn't really add up to me.

user-post.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+echo "Running post build script"
+pip install -r requirements.txt
+npm ci
+
+cd src
+
+python manage.py collectstatic --settings=config.settings.build --noinput
```
CamLamb, line 9: You've moved `collectstatic` out of the Procfile, into this script, but I can't see this script being called?

user-post.sh
WillGibson: We could look at changing this, but it would require some conversation...

https://github.com/uktrade/ci-image-builder/blob/main/build.sh#L82

user-post.sh
WillGibson: In DBT PaaS, `collectstatic` runs when the image is built, because it only needs to happen once per image, not once per container.

src/config/settings/env.py
```diff
@@ -21,16 +21,16 @@
...
-    "ENVIRONMENT": env("APP_ENV"),  # noqa F405
-    "SERVER_TIMEOUT": env("ELASTIC_APM_SERVER_TIMEOUT", default="20s"),  # noqa F405
-}
+#  ELASTIC_APM = {
+#  "SERVICE_NAME": "Digital Workspace",
+#  "SECRET_TOKEN": env("ELASTIC_APM_SECRET_TOKEN"),  # noqa F405
+#  "SERVER_URL": env("ELASTIC_APM_SERVER_URL"),  # noqa F405
+#  "ENVIRONMENT": env("APP_ENV"),  # noqa F405
+#  "SERVER_TIMEOUT": env("ELASTIC_APM_SERVER_TIMEOUT", default="20s"),  # noqa F405
+#  }
```
antroy-madetech, line 34: Need to discuss this on Monday. Do we need to enable this conditionally for non-copilot builds?

src/search/templatetags/search.py
```diff
@@ -44,6 +44,7 @@ def search_category(context, *, category, limit=None, show_heading=False):
     # pages will have the pinned results removed after pagination and cause
     # the pages to have odd lengths.
     search_results = list(pinned_results) + list(search_vector.search(query))
+
```
CamLamb, line 47: Can we revert this change?
We're working on some large search changes so want to keep the changes to this app minimal.

src/config/settings/base.py
```diff
@@ -47,16 +47,14 @@
     AWS_REGION = app_bucket_creds["aws_region"]
     AWS_S3_REGION_NAME = app_bucket_creds["aws_region"]
     AWS_STORAGE_BUCKET_NAME = app_bucket_creds["bucket_name"]
+    AWS_ACCESS_KEY_ID = env("AWS_ACCESS_KEY_ID")
```
CamLamb, line 50: Can you deploy this branch to dev to check that the S3 URL signing still works?

CamLamb (commented):
I've added a couple of comments 😄

src/config/settings/test.py
```diff
@@ -18,10 +17,6 @@
     "AUTO_UPDATE": False,
 }
 
-INSTALLED_APPS += [  # noqa F405
-    "django_extensions",
-]
-
```
marcelkornblum, line 24: would be good to look at the settings inheritance to keep things like this in

CamLamb (approved):
LGTM

PR #465, opened 2023-09-22
DWPF-783 Fixes bug where None was displayed if location in building field was not filled in

nicopicchio:


src/peoplefinder/models.py
```diff
@@ -800,14 +800,13 @@ def get_workdays_display(self) -> str:
...
-                + "<br>"
-                + strip_tags(self.location_in_building)
-            )
         if self.international_building:
             return self.international_building
+        if self.uk_office_location:
+            location_display = self.uk_office_location.name
+            if self.location_in_building:
+                location_display += "<br>" + strip_tags(self.location_in_building)
+            return mark_safe(location_display)  # noqa: S308
```
CamLamb, line 809: Looks good!

Here are some other options that are worth thinking about, I think what you have done is good, just worth considering some of the other options and thinking about pros/cons:

Your way:
```
if self.uk_office_location:
    location_display = self.uk_office_location.name
    if self.location_in_building:
        location_display += "<br>" + strip_tags(self.location_in_building)
    return mark_safe(location_display)
```
Will run `mark_safe` even if it is just the `uk_office_location` that is set in `location_display`. While it isn't an issue, it does mean that if anyone updates the `location_display` with a user entered text field in the future we are open to some funkyness.

Alternatively, you could exit out early by moving the logic around a little:
```

if self.uk_office_location:
    if not self.location_in_building:
        return self.uk_office_location.name
    return mark_safe(
        self.uk_office_location.name + "<br>" + strip_tags(self.location_in_building)
    )
```

This results in the only time we use `mark_safe` is when we are actively adding a string that contains HTML markup `"<br>"` and then stripping the tags from the user provided text field with `strip_tags`.

CamLamb (approved):
LGTM, added a comment with an alternative approach.
I'm happy for this to be merged as is, or with the change, up to you 😄

src/peoplefinder/models.py
nicopicchio: yeah good point. I've refactored with `and not self.location_in_building` rather than nested if though

PR #472, opened 2023-10-04
Wagtail update to 5.1.2

CamLamb:
- [x] Test that the group page permissions are migrated for existing groups


.circleci/config.yml
```diff
@@ -21,6 +21,9 @@ jobs:
       - run:
           name: run containers
           command: docker-compose up -d
+      - run:
+          name: run migrations
+          command: make migrate
```
marcelkornblum, line 26: this _can't_ be necessary, or how on earth has anything run up til now...

src/content/management/commands/create_groups.py
```diff
@@ -26,15 +25,16 @@
...
+    "change_page",
+    "publish_page",
 ]
 
-MODERATOR_PAGE_PERMISSION_TYPES = EDITOR_PAGE_PERMISSION_TYPES + [
-    "lock",
-    "unlock",
+MODERATOR_PAGE_PERMISSIONS = EDITOR_PAGE_PERMISSIONS + [
+    "lock_page",
+    "unlock_page",
```
marcelkornblum, line 37: i take it wagtail ships a migration to move existing permissions over to the new format, right?

src/extended_search/backends/backend.py
```diff
@@ -75,17 +81,53 @@ def _join_and_compile_queries(self, query, fields, boost=1.0):
...
+        if isinstance(field, Field):
+            return field.field_name
+        return field
+
+    def to_field(self, field: Union[str, Field]) -> Field:
+        if isinstance(field, Field):
+            return field
+        return Field(field)
+
+    def backport_fields(self, fields: List[Any]) -> List[str]:
```
CamLamb, line 94: We need to discuss the Wagtail refactor and making sure we are talking about the right type of `fields` at the right layer in the abstraction.

My current understanding:

- ElasticSearch5:
  - remapped_fields: List[str]
  - fields: List[str]
- ElasticSearch6:
  - remapped_fields: List[Field]
  - fields: List[str | Field]
- ElasticSearch7:
  - remapped_fields: List[Field]
  - fields: List[str | Field]
  
My current thoughts are that we should refactor ES6+ to only refer to `Field` objects and then convert them to strings for the calls to `super()` where it is pointing to an ES5 class.

marcelkornblum: I like the approach, but is that in scope here do you think, or do we need to do it upstream and wait for it to go through?

marcelkornblum: closing this - gonna do it upstream

marcelkornblum (approved):

src/content/management/commands/create_groups.py
CamLamb: Just reviewed after our hotfix deploy and it looks good!

PR #473, opened 2023-10-05
Update PII tooling

CamLamb:


rossmiller (changes requested):
Too many files in exclude for my taste! Looks good apart from that.

CamLamb:
I've added the following to pre-commit:
- Yelp detect-secrets
- Black
- Ruff
- Djlint

PR #474, opened 2023-10-05
Clean up search

CamLamb:


marcelkornblum (approved):
we could split out QueryBuilder from Manager and make the Manager accessible from the model in a separate PR, or in this, whichever feels best

CamLamb:
Let's do that in a separate PR so this can be merged in

PR #477, opened 2023-10-06
DWPF-755 Update profile edit to have an "account settings" section

CamLamb:
With an email field that's a dropdown with SSO emails


CamLamb:
This is deployed to dev 😄

marcelkornblum (approved):
I'd be surprised (and disappointed) if that copy stays the same, but technically looks ace

PR #478, opened 2023-10-09
DBTP-480 Infosec 9.4: Password found in configuration file

gabelton:
[Ticket ](https://uktrade.atlassian.net/browse/DBTP-480) arises from recent pen test of new DBT PaaS. Hardcoded variable in docker-compose.yml was identified as a potential security risk, even though it's currently only used in local dev workflow.

N.B. I had to add [an env variable](https://app.circleci.com/settings/project/github/uktrade/digital-workspace-v2/environment-variables?return-to=https%3A%2F%2Fapp.circleci.com%2Fpipelines%2Fgithub%2Fuktrade%2Fdigital-workspace-v2%3Fbranch%3DDPTP-480-Infosec-9.4-Password-found-in-configuration-file) in circle ci, in order to avoid hardcoding the password

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -105,3 +105,6 @@ UK_STAFF_LOCATIONS_DATABASE_URL=psql://postgres:postgres@db:5432/uk_staff_locati
 HIDE_NEWS=False
 
 FEEDBACK_NOTIFICATION_EMAIL_RECIPIENTS="marcel.kornblum@digital.trade.gov.uk,"
+
+# Variables used locally by docker-compose.yml
+DOCKER_COMPOSE_POSTGRES_PASSWORD=postgres
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ common/CACHE
 __pycache__
 !.gitignore
 .venv/
+venv
 .pytest_cache/
 .python-version
 .ruff_cache
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,7 +10,7 @@ services:
       - ./.db:/var/lib/postgresql/data
     environment:
       - POSTGRES_DB=digital_workspace
-      - POSTGRES_PASSWORD=postgres
+      - POSTGRES_PASSWORD=${DOCKER_COMPOSE_POSTGRES_PASSWORD}
     healthcheck:
       test: [ "CMD-SHELL", "pg_isready -U postgres" ]
       interval: 5s


yusufsheiqh (approved):

marcelkornblum:
I'm not sure moving it to an env variable that's then committed to the repo is any better :D 

The password is to a transient data store, which by definition will only hold fixture data also in the repo, or data entered by whoever clones and hosts the project; it's therefore not a data risk in any way.

gabelton:
> I'm not sure moving it to an env variable that's then committed to the repo is any better :D
> 
> The password is to a transient data store, which by definition will only hold fixture data also in the repo, or data entered by whoever clones and hosts the project; it's therefore not a data risk in any way. Note the password is to whichever postgres DB is instantiated to run tests.

I haven't committed the env variable, only an example of what it might look like. 

I know that the risk is small and that it's only used in build / test workflow, but I guess the concern is if someone were to use the docker-compose file in another context and forget to remove the hardcoded variable. I know this PR introduces a dependency and means storing a variable in circleci, but at least that's protected and limited to those with project access.

marcelkornblum:
> I haven't committed the env variable, only an example of what it might look like.
> 
> I know that the risk is small and that it's only used in build / test workflow, but I guess the concern is if someone were to use the docker-compose file in another context and forget to remove the hardcoded variable. I know this PR introduces a dependency and means storing a variable in circleci, but at least that's protected and limited to those with project access.

I'd actually argue the risk isn't small, it's zero. If someone uses the hardcoded variable in another context, it still doesn't introduce risk since nobody has any restricted data in a docker container or an environment used with docker compose. If they did, _that_ would be a security risk to be addressed.

Why not get the secret used in tests populated at runtime from aws secrets? Then it could be encrypted and the risk mitigated even more. The reason is because the inherent risk profile doen't justify the extra work, overhead and maintenance. The same applies to the changes in this PR; to follow the logic that _any_ risk needs mitigation we could spend all our time introducing levels of indirection.

marcelkornblum:
Just to add that my understanding is that adding environment variables to CircleCI to run tests is considered bad practice at the department, since it encourages people to code in a less secure way. The argument is that once secrets are required it erodes the overall "code in the open with a security mindset" approach and adds mental load that is shown to reduce security.

yusufsheiqh:
Although I approved this PR, I agree with @marcelkornblum - if we can pushback and close the ticket as we know there is no risk involved, I think that would be better. This seems like a tick in the box thing to do which seems unnecessary.

gabelton:
> Why not get the secret used in tests populated at runtime from aws secrets? Then it could be encrypted and the risk mitigated even more. 

My understanding is that we would do this when 

> Although I approved this PR, I agree with @marcelkornblum - if we can pushback and close the ticket as we know there is no risk involved, I think that would be better. This seems like a tick in the box thing to do which seems unnecessary.

The [original pen test description](https://dbis-my.sharepoint.com/personal/maureen_kaggwa_businessandtrade_gov_uk/_layouts/15/onedrive.aspx?id=%2Fpersonal%2Fmaureen%5Fkaggwa%5Fbusinessandtrade%5Fgov%5Fuk%2FDocuments%2FDBT%20PaaS%20Reports%2FPRISMINFOSEC%2DDBT%2DNCSC%5FCHECK%2DPAAS%5FAWS%5FAND%5FAPPLICATION%5FCONFIG%5FREVIEW%2DSECURITY%5FASSESSMENT%2DREPORT%2D10608%2Epdf&parent=%2Fpersonal%2Fmaureen%5Fkaggwa%5Fbusinessandtrade%5Fgov%5Fuk%2FDocuments%2FDBT%20PaaS%20Reports) acknowledges that as well, so I'm not sure why they decided to raise it still

PR #480, opened 2023-10-13
DWPF-749 Content owners

CamLamb:


marcelkornblum (approved):
feels like we've done this a few different ways across a few different codebases. this looks nice but i wish we had a unified approach...

PR #482, opened 2023-10-19
WIP: Clean user records

marcelkornblum:
made a management command that will:
- delete all user records with no associated profile
- delete all profiles with no associated user

made another management command that will:
- mark users as inactive where their profile is marked as inactive
- mark profiles as inactive where the user is marked as inactive
- mark users/profiles as inactive when their last login is over 90 days ago (configurable), and they have an inactive profile on SSO (NB dependent on a PR on staff-sso) (NB is API intensive and so can be batched)

Amended the active/inactive setters in the codebase (inc admin) to make sure they stay in sync

### DIFF
--- a/src/peoplefinder/models.py
+++ b/src/peoplefinder/models.py
@@ -783,6 +783,10 @@ class Person(Indexed, models.Model):
         from peoplefinder.services.person import PersonService
 
         self.profile_completion = PersonService().get_profile_completion(person=self)
+
+        if self.user and self.is_active != self.user.is_active:
+            self.user.is_active = self.is_active
+            self.user.save()
         return super().save(*args, **kwargs)
 
     def get_first_name_display(self) -> str:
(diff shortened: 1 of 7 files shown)


src/peoplefinder/models.py
```diff
@@ -783,6 +783,10 @@ def save(self, *args, **kwargs):
         from peoplefinder.services.person import PersonService
 
         self.profile_completion = PersonService().get_profile_completion(person=self)
+
+        if self.user and self.is_active != self.user.is_active:
+            self.user.is_active = self.is_active
```
CamLamb, line 788: If the profile is set to active and the user is set to inactive, it will update the user to be active, is this the desired behaviour?

Is there any way that updates the User is active flag? (maybe from SSO?)

marcelkornblum: I put this as provocation - my assumption is whichever record (User / Person) is being saved, you _probably_ want the other record's is_active to match.... ?

marcelkornblum: the sso connection is waiting on a erlated PR but is there in the management command; I could abstract this to a service

CamLamb:
Closing as this is an old branch/PR and it is being addressed in some future work

PR #484, opened 2023-10-25
Allow per-model field overrides

marcelkornblum:
ticket 548

Indexed models can override IndexedField definitions in order to e.g. change the analysis types, boosting, etc.

- [x] Update main query generator entrypoint
- [x] add new Filtered SearchQuery type, with accompanying compiler and mapping updates
- [x] restructure IndexManagers to be inner classes of Indexed models
- [x] add to Indexed and querybuilder to handle deliberate direct field overrides without hurting underlying functionality
- [ ] tests:
  - [x] fix newly broken tests
  - [x] test new additional functionality
  - [ ] test that subclass overrides stop superclass extra fields being created (i.e. explicit vs tokenized, etc)
  - [x] test that subclass overrides with extra fields are indexed  (i.e. explicit vs tokenized, etc)
  - [x] test that subclass overrides boosting
  - [x] test that settings overriding subclassed overrides work as intended
  - [x] verify that the new filtered search does in fact filter correctly on "all pages" search


src/extended_search/backends/backend.py
```diff
@@ -231,6 +239,42 @@ def _compile_nested_query(self, query, fields, boost=1.0):
...
+
+        return {
+            "bool": {
+                "must": self._join_and_compile_queries(query.subquery, fields, boost),
+                "filter": compiled_filters,
+            }
+        }
+
+    def _process_lookup(self, field, lookup, value):
+        if type(field) == str:
```
marcelkornblum, line 264: why do we need this? self.mapping.get_field_column_name(field) should handle it I tihnk?

src/extended_search/managers/index.py
```diff
@@ -98,9 +98,28 @@ def get_mapping(cls):
 
     @classmethod
     def get_search_fields(cls):
-        index_fields = []
+        cls.generated_fields = []
```
CamLamb, line 101:
```
cls.generated_fields = [
    cls._get_search_fields_from_mapping(field_mapping)
    for field_mapping in cls.get_mapping()
]
```

CamLamb (approved):
LGTM

PR #486, opened 2023-10-25
Fix e2e testrunner

marcelkornblum:
* Bumped Debian to bookworm
* Bumped playwright explicitly since it got forced to an old version somehow
* Moved node installation to newer approach


CamLamb (approved):
LGTM

PR #489, opened 2023-10-26
Fix preview

CamLamb:


pyproject.toml
```diff
@@ -14,7 +14,7 @@ django-crispy-forms = "^1.14.0"
 django-environ = "^0.10.0"
 # django_redis = "^5.2.0"
 django-settings-export = "^1.2.1"
-django-simple-history = "^3.3.0"
+django-simple-history = { git = "https://github.com/jazzband/django-simple-history.git", rev = "29108a5" }
```
CamLamb, line 17: django-simple-history needs a new release to fix this issue, in the meantime we can pin a commit.

For added protection, we should fork the repo and point this at the head on our own fork

marcelkornblum: agree with the forking approach. annoying moving to a git ref :(

what's the actual issue?

CamLamb: https://github.com/jazzband/django-simple-history/pull/1256
TLDR: they removed a condition that was thought to be unnecessary, but was actually needed in some scenarios (Wagtail Preview managed to hit this scenario)

marcelkornblum (approved):

PR #491, opened 2023-11-01
DWPF-862 Prevent using javascript as a url in team description markdown

CamLamb:


src/peoplefinder/templatetags/markdown.py
```diff
@@ -3,12 +3,15 @@
...
 register = template.Library()
 
 
 @register.filter(needs_autoescape=True)
-def markdown(text, autoescape=True):
+def markdown(text: str, autoescape=True):
+    # Remove any javascript: or data: from the text to prevent XSS
+    text = text.replace("javascript:", "")
+    text = text.replace("data:", "")
+
```
marcelkornblum, line 14: maybe also remove script tags?

CamLamb: We don't need to worry about HTML tags as those are escaped already:
Screenshot 2023-11-01 at 13 13 29

marcelkornblum (approved):

PR #493, opened 2023-11-03
Search score functions for proximity

marcelkornblum:
ticket 605


src/content/models.py
```diff
@@ -291,7 +291,12 @@ class IndexManager(ModelIndexManager):
                 explicit=True,
             ),
             IndexedField("is_creatable", filter=True),
-            IndexedField("published_date", proximity=True),
+            ScoreFunction(
```
CamLamb, line 294: I love how simple this implementation level is 😄

marcelkornblum: Ideally it wouldn't be in "fields" but that seemed like too big a change... thoughts?

marcelkornblum:
closing in favour of #581

PR #503, opened 2023-11-14
DBTP-586 Setup DBT Platform build configuration

acodeninja:
- Remove the old-style build configuration files.
- Add the new build configuration file `.copilot/config.yml`.
- Remove `pip` and `npm` install from post build script as not needed.


.copilot/image_build_run.sh
```diff
@@ -5,12 +5,6 @@ set -e
 
 echo "Running post build script"
 
-echo "Running pip install"
-pip install -r requirements.txt
```
marcelkornblum, line 9: do these just happen "magically" ?

I'm a big fan of the file name change BTW :)

marcelkornblum (commented):
what there is looks great but there's so little - where does that config now live?

.copilot/image_build_run.sh
acodeninja: Yeah the buildpack `pip-install` takes care of this, then the script is run in [this context](https://github.com/uktrade/ci-image-builder/blob/main/image_builder/codebase/codebase.py#L43C1-L55C10)

marcelkornblum: what about the npm one? 

This sort of thing being implicit (rather than explicit) is prettier but I'm always suspicious of magic :) 

e.g. we really use the pyproject.toml config but output to requirements with a script; would be nice to move across to using pyproject.toml directly but how would we know if that's doable...

acodeninja: So you can run the build tool locally though it's not great at the moment, and we will be integrating it into `copilot-tools` to sit alongside the `copilot-helper` command.

Currently we use the the language detection model [here](https://github.com/uktrade/ci-image-builder/blob/main/image_builder/codebase/language/__init__.py) which will pick up python and nodejs codebases.

Then when paketo goes through it's detection phase it will decide which buildpacks to use. For example:

For NodeJS

- nodejs buildpack is included because there is a `package.json` file present
- paketo then works out which buildpacks to apply
  - the version of nodejs to use is detected based on the `pakcage.json` property `engines` if present, otherwise it will install the latest lts via the buildpack `paketo-buildpacks/node-engine` is included.
  - the `package.json` has dependencies present, so the `paketo-buildpacks/npm-install` buildpack is applied.
  - the presence of a `build` entry in the `package.json` `scripts` property leads to the `paketo-buildpacks/node-run-script` buildpack being added and the build script being run.

For Python

- the python buildpack is included because a `requirements.txt`, `pyproject.toml` or `runtime.txt` file with `python` in it was present.
- `pyproject.toml` and `runtime.txt` are used to decide which version of python is needed.
- paketo then works out which buildpacks to apply
  - `paketo-buildpacks/cpython` is applied, because python
  - `paketo-buildpacks/pip` and `paketo-buildpacks/pip-install` are applied because there is a requirements.txt file.

Notes
- Currently we don't support `poetry` only codebases, but this will come in the future and the only change to make in the codebase is to delete the `requirements.txt` file.

marcelkornblum (approved):

codebuild/process.yml
```diff
@@ -1,12 +0,0 @@
...
-  process:
-    # The way DBT PaaS works currently we need one entry
-    # here for each of the entries in the Procfile.
-    # The only actual different in the three images created
-    # is the process they start up with.
-    # There is a plan ensure it only needs to build one
-    # image in the near future.
-    - web
-    - beat # celery beat
-    - worker # celery worker
```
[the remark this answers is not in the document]
acodeninja: These are already defined in your `Procfile` so we removed the need for duplication.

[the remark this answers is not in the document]
acodeninja: The `Procfile` is read in and processed by [this class](https://github.com/uktrade/ci-image-builder/blob/main/image_builder/codebase/processes.py)

PR #505, opened 2023-11-14
DWPF-788 Fix searching for people on office location

SamDudley:


marcelkornblum (approved):
so happy to have test locations!

PR #506, opened 2023-11-14
DWPF-778 Show bad results message

nicopicchio:


src/search/search.py
```diff
@@ -7,20 +7,16 @@
 
 
 class SearchVector:
-    def __init__(self, request, annotate_score=False):
+    def __init__(self, request):
         self.request = request
-        self.annotate_score = annotate_score
+        self.annotate_score = True
```
CamLamb, line 12: `self.annotate_score` is no longer used anywhere

src/search/utils.py
```diff
@@ -136,3 +142,86 @@ def split_query(query: str) -> list[str]:
...
+    boost_values = set()
+    subqueries = get_all_subqueries(query)
+
+    for subquery in subqueries[category]:
+        boost_values.add(round(subquery["boost"], 2))
+
+    return statistics.median(boost_values) * bad_score_threshold_multiplier
+
+
+def query_has_bad_results(query, category, pinned_results, search_results):
```
CamLamb, line 218: Can you add a comment to explain the purpose of this method?

src/search/utils.py
```diff
@@ -136,3 +142,86 @@ def split_query(query: str) -> list[str]:
...
+            query, Person, mapping
+        )
+        get_query_info(subqueries["people"], field, mapping, analyzer_field_suffices)
+    for mapping in Team.IndexManager.get_mapping():
+        field = Team.IndexManager._get_search_query_from_mapping(query, Team, mapping)
+        get_query_info(subqueries["teams"], field, mapping, analyzer_field_suffices)
+    return subqueries
+
+
+def get_bad_score_threshold(query, category):
```
CamLamb, line 207: Can you add a comment to explain the purpose of this method?

src/search/templates/search/partials/search_results_category.html
```diff
@@ -1,4 +1,5 @@
 <div class="search-results__category">
+    {% if show_bad_results_message %}<h1>These results might not be what you are looking for</h1>{% endif %}
```
marcelkornblum, line 2: I think we need a few different versions of this copy depending on other criteria - are there lots of results or only a few? which type of search are we on?

src/search/utils.py
marcelkornblum, line 218: could we rename this to be more on-the-nose please? maybe `only_has_bad_results` or `has_no_good_results` or something?

src/search/utils.py
```diff
@@ -1,6 +1,12 @@
 import re
 import unicodedata
+import statistics
```
marcelkornblum, line 3: i tihnk this is overkill for what we want to do

src/search/utils.py
```diff
@@ -136,3 +142,86 @@ def split_query(query: str) -> list[str]:
...
+        )
+        get_query_info(subqueries["people"], field, mapping, analyzer_field_suffices)
+    for mapping in Team.IndexManager.get_mapping():
+        field = Team.IndexManager._get_search_query_from_mapping(query, Team, mapping)
+        get_query_info(subqueries["teams"], field, mapping, analyzer_field_suffices)
+    return subqueries
+
+
+def get_bad_score_threshold(query, category):
+    bad_score_threshold_multiplier = 1
```
marcelkornblum, line 208: this should be extracted to an easier-to-change setting. I tihnk an ENV var in settings is probably best.

src/search/utils.py
```diff
@@ -136,3 +142,86 @@ def split_query(query: str) -> list[str]:
...
+    boost_values = set()
+    subqueries = get_all_subqueries(query)
+
+    for subquery in subqueries[category]:
+        boost_values.add(round(subquery["boost"], 2))
+
+    return statistics.median(boost_values) * bad_score_threshold_multiplier
```
marcelkornblum, line 215: i am not convinced median is right - more like the 70th percentile maybe, just as the curve gets steeper? either way I don't think we need the stats package here; all we're doing is choosing the nth value where N / total = {our percentile}. so if we want median we total the number of boosts, halve the total to get N and select the boost at position N in the list.

src/search/utils.py
```diff
@@ -136,3 +142,86 @@ def split_query(query: str) -> list[str]:
         parts.append(group)
 
     return parts
+
+
+def get_query_info(fields, field, mapping, suffix_map):
```
marcelkornblum, line 147: these copied-acrss functions would ideally get tidied up, they're such a mess. not going to mind if you don't though :)

marcelkornblum (changes requested):
Good as far as it goes, but there are a few bits missing from what the ticket needs (see specific comments)

src/search/templates/search/partials/search_results_category.html
nicopicchio: we do but I am waiting for Bobby to come up with a final copy. initially it was `hello world`, I've only changed it to give a more meaningful message before opening this PR

marcelkornblum: sure, but it also needs to be more than one message depending on other factors - so those factors ought to make an appearance here

src/search/utils.py
nicopicchio: ok, what do you suggest instead?

marcelkornblum: commented below

src/search/utils.py
nicopicchio: renamed to `has_only_bad_results`

src/search/utils.py
nicopicchio: done

src/search/templatetags/search.py
```diff
@@ -33,19 +34,26 @@
...
 )
 def search_category(
-    context, *, category, tab_name=None, limit=None, show_heading=False
+    context,
+    *,
+    category,
+    tab_name=None,
+    limit=None,
+    show_heading=False,
+    show_bad_threshold_message=True,
```
marcelkornblum, line 43: could we rationalise the variables `show_bad_threshold_message` and `show_bad_results_message` across the whole codebase please? I can't see a reason to have them being different and it'd be easier to understand the code if they were named the same (assuming they are in fact doing the same thing)

marcelkornblum (commented):
just one last thing please - otherwise looking fantastic

src/search/templatetags/search.py
nicopicchio: renamed `show_bad_results_message`

PR #507, opened 2023-11-15
Nested support and field naming consistency

marcelkornblum:
allow querybuilder to access Nested ES fields
refactor missed indexmanager logic that should have been removed previously

### DIFF
--- a/src/extended_search/index.py
+++ b/src/extended_search/index.py
@@ -42,85 +42,17 @@ class Indexed(index.Indexed):
     indexed_fields = []
 
     @classmethod
-    def _get_analyzer_name(cls, analyzer_type):
-        from extended_search.settings import extended_search_settings as search_settings
-
-        analyzer_settings = search_settings["analyzers"][analyzer_type.value]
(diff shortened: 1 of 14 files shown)


src/extended_search/index.py
```diff
@@ -382,14 +423,62 @@ def __init__(
...
+        ]
+
+    # def get_filter_field_variants(self):
+    #     return super().get_filter_field_variants()
+
+    # def get_autocomplete_field_variants(self):
+    #     return super().get_autocomplete_field_variants()
+
+
+# class MultiQueryRelatedField(RelatedField):
```
CamLamb, line 473: This can be removed now yea?

CamLamb (approved):

src/extended_search/index.py
marcelkornblum: I tihnk that's the way we went.... ?

CamLamb:
Merged into https://github.com/uktrade/digital-workspace-v2/pull/497

PR #509, opened 2023-11-16
DWPF-784 Typeahead strong match & no results redirect

0RWilliams:
- Added logic around zero results tabs, so that when no results are present in the tab, redirect.
- Added Logic & UI for autocompletion search functionality via search-form.

### DIFF
--- a/src/search/search.py
+++ b/src/search/search.py
@@ -9,25 +9,19 @@ from working_at_dit.models import PoliciesAndGuidanceHome
 
 
 class SearchVector:
-    def __init__(self, request, annotate_score=True):
+    def __init__(self, request):
         self.request = request
-        self.annotate_score = annotate_score
 
     def _wagtail_search(self, queryset, query, *args, **kwargs):
         """
         Allows e.g. score annotation without polluting overridden search method
         """
-        return_method = queryset.search(query, *args, **kwargs)
-
-        if self.annotate_score:
-            return_method = return_method.annotate_score("_score")
+        return_method = queryset.search(query, *args, **kwargs).annotate_score("_score")
 
         return return_method
 
     def _wagtail_autocomplete(self, queryset, query, *args, **kwargs):
-        return_method = queryset.autocomplete(query, *args, **kwargs)
-
-        return return_method
+        return queryset.autocomplete(query, *args, **kwargs)
 
     def get_queryset(self):
         raise NotImplementedError
--- a/src/search/search.py
+++ b/src/search/search.py
@@ -23,10 +23,10 @@ class SearchVector:
             return_method = return_method.annotate_score("_score")
 
         return return_method
-    
-    def _wagtail_autocomplete(self,queryset,query,*args, **kwargs):
-        return_method = queryset.autocomplete(query,*args,**kwargs)
-        
+
+    def _wagtail_autocomplete(self, queryset, query, *args, **kwargs):
+        return_method = queryset.autocomplete(query, *args, **kwargs)
+
         return return_method
 
     def get_queryset(self):
@@ -56,10 +56,10 @@ class PagesSearchVector(SearchVector):
 
     def pinned(self, query):
         return self.get_queryset().pinned(query)
-    
+
     def autocomplete(self, query, *args, **kwargs):
         queryset = self.get_queryset().not_pinned(query)
-        return self._wagtail_autocomplete(queryset,query,*args, **kwargs)
+        return self._wagtail_autocomplete(queryset, query, *args, **kwargs)
 
     def search(self, query, *args, **kwargs):
         queryset = self.get_queryset().not_pinned(query)
@@ -117,7 +117,7 @@ class PeopleSearchVector(SearchVector):
         return self._wagtail_search(queryset, query, *args, **kwargs)
 
     def autocomplete(self, query, *args, **kwargs):
-        return self._wagtail_autocomplete(self.get_queryset(),query,*args, **kwargs)
+        return self._wagtail_autocomplete(self.get_queryset(), query, *args, **kwargs)
 
 
 class TeamsSearchVector(SearchVector):
@@ -130,4 +130,4 @@ class TeamsSearchVector(SearchVector):
         return self._wagtail_search(queryset, query, *args, **kwargs)
 
     def autocomplete(self, query, *args, **kwargs):
-        return self._wagtail_autocomplete(self.get_queryset(),query,*args, **kwargs)
+        return self._wagtail_autocomplete(self.get_queryset(), query, *args, **kwargs)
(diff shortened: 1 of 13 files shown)


src/search/templates/search/partials/result/autocomplete_page.html
```diff
@@ -0,0 +1,73 @@
+{% load webpack_static from webpack_loader %}
```
marcelkornblum, line 1: do we still need this page? if it was just for testing let's delete it please...

src/search/templatetags/search.py
```diff
@@ -78,6 +82,55 @@ def search_category(context, *, category, limit=None, show_heading=False):
     }
 
 
+# Method for querying using wagtails default autocomplete functionality
+#
+def autocomplete(request, query, all_results=False):
+    limit = 3
+    search_results = {}
+
+    if all_results:
```
marcelkornblum, line 91: so the typeahead shows different results depending which page / search tab you're on? That feels a bit off to me, I think it probably should always do all results...

src/search/templates/search/partials/result/autocomplete_page.html
marcelkornblum: or is this the typeahead rendered results?

marcelkornblum (commented):
mostly great, a couple of specific comments.
also would be good to split things into smaller PRs (one for redirect, one for typeahead) and to "request reviews" from people  so they get notifications :)

src/search/templates/search/partials/result/autocomplete_page.html
0RWilliams: Needed for when the person image is not present, i.e. showing the no-photo img.

src/search/templatetags/search.py
0RWilliams: Not quite, it's based on a Bool value "all_results", which is defaulted to False, i put it in, so i could switch between 3x3x3 or YxYxY, can remove it to be honest.

marcelkornblum: i reckon remove it if not needed please just to keep it simple

src/search/templates/search/partials/result/autocomplete_page.html
marcelkornblum: sorry i meant the whole file - but i think it's the partial that's rendered in response to the autocomplete right?

if so then all good.

src/search/search.py
```diff
@@ -9,7 +9,7 @@
 
 
 class SearchVector:
-    def __init__(self, request, annotate_score=False):
+    def __init__(self, request, annotate_score=True):
         self.request = request
         self.annotate_score = annotate_score
```
CamLamb, line 14: We can just remove `annotate_score` from the init kwargs, and `self.annotate_score` from the `SearchVector`

CamLamb: Nico's work does this, so there's little to no harm in removing it here too. (don't forget to drop the `if self.annotate_score:` condition below)

src/search/search.py
```diff
@@ -24,6 +24,11 @@ def _wagtail_search(self, queryset, query, *args, **kwargs):
 
         return return_method
 
+    def _wagtail_autocomplete(self, queryset, query, *args, **kwargs):
+        return_method = queryset.autocomplete(query, *args, **kwargs)
+
+        return return_method
```
CamLamb, line 30: This can be updated to just:
`return queryset.autocomplete(query, *args, **kwargs)`

src/search/templates/search/partials/search_results_all.html
```diff
@@ -2,10 +2,10 @@
 
 <div class="search-results__all">
     <div class="left-col">
-        <div class="pages">{% search_category category='all_pages' show_heading=True %}</div>
+        <div class="pages">{% search_category category='all_pages' tab_name='all' show_heading=True %}</div>
     </div>
     <div class="right-col">
-        <div class="people">{% search_category category='people' limit=3 show_heading=True %}</div>
-        <div class="teams">{% search_category category='teams' limit=3 show_heading=True %}</div>
+        <div class="people">{% search_category category='people' tab_name='all' limit=3 show_heading=True %}</div>
```
CamLamb, line 8: `tab_name='all'` feels a little confusing in this context.
The name suggests that the People category is the all tab, maybe we can consider a different name?

src/search/templates/search/partials/search_results_all.html
```diff
@@ -2,10 +2,10 @@
...
+        <div class="teams">{% search_category category='teams' tab_name='all' limit=3 show_heading=True %}</div>
```
CamLamb, line 9: `tab_name='all' similar comment to the line above.

src/search/templatetags/search.py
```diff
@@ -78,6 +82,55 @@ def search_category(context, *, category, limit=None, show_heading=False):
     }
 
 
+# Method for querying using wagtails default autocomplete functionality
+#
+def autocomplete(request, query, all_results=False):
```
CamLamb: Since this isn't a templatetag, we should remove this out of this file.
It feels more like a util method

src/search/templates/search/partials/result/autocomplete_page.html
0RWilliams: Yes, that's the case.

src/search/templatetags/search.py
0RWilliams: Removed.

src/search/templates/search/partials/search_results_all.html
0RWilliams: Do you have a suggestion for what name to use instead?
The name represents the fact that it's on the 'all pages' view, and not an individual view.

src/search/templatetags/search.py
0RWilliams: I have moved this to utils.

marcelkornblum: I agree it should be moved, but this ends up having a knock-on effect to the other search methods and results in circular imports etc. 
I think we should instead rationalise all this and the template structure all in one go; for now just keep it as another templatetag for consistency

src/search/templates/search/partials/search_results_all.html
marcelkornblum: @CamLamb hard to think of a way to improve this from my PoV - it's a knock-on effect of the template structure we have that the code needs ot know which "view" we're in as well as which search type we're doing. I'd probably defer this to a wider refactor personally and hopefully we could sort all the naming out along with structure

marcelkornblum (approved):
LGTM!

src/search/templates/search/partials/search_results_all.html
CamLamb: Yea, this isn;t a blocker for now, and we can change it later if needed :)

PR #511, opened 2023-11-20
DWPF-786 Show recently inactive profiles in search results to all users

SamDudley:


CamLamb (approved):
LGTM from an initial look 😄

marcelkornblum (approved):
Just a Q about the new field, otherwise good

PR #531, opened 2023-12-18
Load testing configuration

CamLamb:


CamLamb:
@SamDudley & @marcelkornblum 
Are we happy for me to merge this PR?

marcelkornblum:
I guess the reason we haven't is it's not tied in to anything (like a CI event, etc). I guess we can merge it but I would rather have a way it runs / what have you first...

PR #533, opened 2023-12-20
Add asim formatter (pt 2)

gabelton:
Latest image is deployed in[ ECR](https://eu-west-2.console.aws.amazon.com/ecr/repositories/private/854321987474/intranet/digital-workspace?region=eu-west-2), but not seeing ASIM logs coming through. Don't believe build settings are being used in the container when I exec into it. Hoping this change should fix


src/config/settings/base.py
```diff
@@ -578,6 +578,9 @@
     },
 }
 
+if is_copilot():
+    LOGGING["handlers"]["ecs"]["formatter"] = "asim_formatter"
```
WillGibson, line 582: I am not certain, but I think you might ned to add the ASIM handler to the relevant loggers, like this...

https://github.com/uktrade/demodjango/blob/main/demodjango/settings.py#L67

WillGibson: I think then the logs will come out in both formats and Cyber can just ingest the ASIM formatted ones.

gabelton: That happens [here](https://github.com/uktrade/digital-workspace-v2/blob/f7d8cfa3e521cb32950e6125aeb4e2d57b452c30/src/config/settings/base.py#L534C22-L534C22)

WillGibson: I meant add the ASIM formatter as well as the ECS one. There may be some Digital Workspace context I've not got though, so I might be misleading you.

gabelton: I think DW only need the ASIM one in the new platform. This is just a stopgap, in case we need to rollback to PaaS and want the ecs formatted logs again. Once the new site is stable, at some point we'll want to remove the ECS formatter from dw entirely

WillGibson (approved):

PR #535, opened 2023-12-21
Fix homepage tweets cache miss

SamDudley:
Co-authored-by: Cameron Lamb <cameron.lamb@digital.trade.gov.uk>
Co-authored-by: Marcel Kornblum <marcel.kornblum@digital.trade.gov.uk>


CamLamb (approved):
LGTM!

In a world where the twitter integration works, `if tweets is None:` should be `if not tweets:`

PR #542, opened 2024-01-30
Fix the end to end tests

CamLamb:
Update to use nodejs 20


SamDudley (approved):

CamLamb:
For some reason, the main menu isn't being created in the test even though it looks like it should be.

SamDudley (approved):

PR #552, opened 2024-03-14
DWPF-1472 Removal of site alert banner

nicopicchio:


alsyx (commented):
Is the banner with the info text going to be light blue by default when this alert banner is removed? @nicopicchio

PR #555, opened 2024-03-20
Remove GovPaas holdovers

marcelkornblum:


src/config/settings/base.py
```diff
@@ -438,14 +420,7 @@
...
 # Redis
-if "redis" in VCAP_SERVICES:
-    credentials = VCAP_SERVICES["redis"][0]["credentials"]
-    CELERY_BROKER_URL = "rediss://:{0}@{1}:{2}/0?ssl_cert_reqs=required".format(
-        credentials["password"],
-        credentials["host"],
-        credentials["port"],
-    )
-elif is_copilot():
+if is_copilot():
```
alsyx, line 423: is this function from here: https://github.com/uktrade/copilot-python/blob/main/dbt_copilot_python/utility.py
I thought it would return a boolean but it returns the COPILOT_ENVIRONMENT_NAME, how does it work? Does it default to false if it is not a copilot environment and true if it returns an environment name? @marcelkornblum

marcelkornblum: I've not actually looked at it before, but yes it's that function. It does actually return a boolean:
```
return "COPILOT_ENVIRONMENT_NAME" in os.environ
```
the `in` means that expression will evaluate to either True or False, based on whether the `COPILOT_ENVIRONMENT_NAME` ENV var is set or not.

CamLamb (approved):
LGTM, this should be deployed to dev to ensure all is still working before this PR can be merged

SamDudley (approved):
LGTM!

PR #556, opened 2024-03-22
DWPF-1504 Update DW Setup Steps in Readme

alsyx:


nicopicchio:
I wonder if we can simplify this even more by sticking`npm install` and `npm run build` inside `make setup`?

nicopicchio:
In fact, `npm install` runs with `make local-setup` so we could just add `npm run build` like we do in [IRAP](https://github.com/uktrade/irap/blob/720f9a2ba239a80c5741f23005f08a918bb3888d/Makefile#L36)

CamLamb (approved):

PR #557, opened 2024-03-25
DWPF-1516 Minor cookie banner restyling

nicopicchio:


assets/stylesheets/components/_site_alert.scss
```diff
@@ -8,6 +8,7 @@
     color: white;
     a, a:visited, a:link {
       color: white;
+      font-size: 1rem;
```
marcelkornblum, line 11: would it not work to instead make the class `govuk-body-s` in the banner?

assets/stylesheets/govuk-overrides/main.scss
```diff
@@ -20,6 +20,7 @@
 
 .govuk-phase-banner {
   width: 100%;
+  border-bottom: none;
```
marcelkornblum, line 23: is it not the cookie banner with the border?

nicopicchio: no the cookie-banner is an external wrapper. what has the border is the phase banner which inside it https://github.com/uktrade/digital-workspace-v2/blob/6cf86a32444da0d7efa2153785f964e5867e9ddf/src/core/templates/tags/cookie_banner.html#L6

assets/stylesheets/components/_site_alert.scss
nicopicchio: yeah good point, sorted it

marcelkornblum (approved):
merci!

PR #568, opened 2024-04-04
Add user groups report

SamDudley:
- **Add user groups report**
- **Fix import sorting**

### DIFF
--- a/src/core/views.py
+++ b/src/core/views.py
@@ -1,11 +1,17 @@
+import csv
 import logging
 
 from django.conf import settings
+from django.contrib.postgres.aggregates import StringAgg
+from django.core.exceptions import PermissionDenied
+from django.http import HttpRequest, HttpResponse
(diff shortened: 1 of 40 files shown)


src/core/views.py
```diff
@@ -86,3 +92,31 @@ def page_problem_found(request):
         "core/page_problem_found.html",
         {"form": form, "message_sent": message_sent},
     )
+
+
+@require_GET
+def user_groups_report(request: HttpRequest, *args, **kwargs) -> HttpResponse:
+    if not request.user.is_superuser:
```
CamLamb, line 99: is everyone that needs access a superuser?

CamLamb (approved):

src/core/views.py
SamDudley: no, but I suggest we look into the correct permission checks and UI afterwards

PR #581, opened 2024-04-22
DWPF 564 Search Proximity By Numerical Score

0RWilliams:
Adds a new ScoreFunction Index to affect query scores when comparing against a numerical value, such as profile completion.

- [x] Tests are up to date


src/extended_search/tests/test_scorefunction.py
```diff
@@ -0,0 +1,91 @@
...
+    CustomSearchVector.model = StandardIndexedModelWithScoreFunctionOriginFifty
+    results = CustomSearchVector(request).search("title")
+
+    # Check the order of the objects
+    assert len(results) == 2
+    assert results[0].pk == obj2.pk
+    assert results[1].pk == obj1.pk
+
+
+def test_inherited_scorefunction_closer_to_zero_is_better(db, mocker):
```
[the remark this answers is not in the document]
CamLamb: Something odd is happening with the indexing of the filter... might need to work on this a bit tbh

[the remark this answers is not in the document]
CamLamb: What is odd is that this test is trying to copy what is happening for on the NewsPage:
```
ScoreFunction(
    "gauss",
    field_name="last_published_at",
    scale="365d",
    decay=0.3,
)
```

This ScoreFunction is defined on the NewsPage model, but the `last_published_at` field is defined on one of it's inherited classes. It gets indexed as expected. So it is strange that this works, but the test that is trying to recreate this fails...

src/peoplefinder/models.py
```diff
@@ -684,10 +684,13 @@ class RemoteWorking(models.TextChoices):
...
-            "profile_completion",
-            proximity=True,
-            boost=2.0,
+        ScoreFunction(
+            "linear",
+            field_name="profile_completion",
+            origin=100,
+            offset=5,
+            scale=50,
+            decay=0.5,
```
marcelkornblum, line 693: i feel like this should be smaller - maybe 0.1 ?

marcelkornblum: also scale = 0

src/news/models.py
```diff
@@ -156,6 +157,12 @@ def search_categories(self):
             "pinned_on_home",
             filter=True,
         ),
+        ScoreFunction(
+            "gauss",
+            field_name="last_published_at",
+            scale="365d",
+            decay=0.3,
```
CamLamb, line 164: Based on this image:
image
Is dropping the score to 30% by 1 year the desired outcome?

src/peoplefinder/models.py
```diff
@@ -690,10 +690,13 @@ class RemoteWorking(models.TextChoices):
...
-            "profile_completion",
-            proximity=True,
-            boost=2.0,
+        ScoreFunction(
+            "linear",
+            field_name="profile_completion",
+            origin=100,
+            offset=5,
+            scale=50,
+            decay=0.05,
```
CamLamb, line 699: Based on this image:
image

I feel like the decay should actually be 0.95 so that at 50% profile completion, 5% of the score will have been lost

CamLamb: I've updated this to 0.95 for now, but we can discuss it further if needed

src/news/models.py
CamLamb: Note: Add an offset of 2 weeks as news should stay relevant for a while

CamLamb (approved):

PR #592, opened 2024-05-07
DWPF-1673 Signposting link for pages

nicopicchio:
- [x] Tests are up to date

### DIFF
--- a/src/working_at_dit/templates/working_at_dit/working_at_dit_home.html
+++ b/src/working_at_dit/templates/working_at_dit/working_at_dit_home.html
@@ -8,6 +8,9 @@
     <div class="govuk-grid-row">
         <div class="govuk-grid-column-two-thirds">
             <h1 class="govuk-heading-l">{{ page.title }}</h1>
+            {% comment %} {% if page.is_published %}
+                <a href="{% url 'page_name_here' page.id%}" class="govuk-button govuk-button--secondary">Check out the new HR Service Page</a>
+            {% endif %} {% endcomment %}
             {% for block in page.body %}
                 {% include_block block %}
             {% endfor %}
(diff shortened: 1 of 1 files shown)


src/working_at_dit/templates/working_at_dit/working_at_dit_home.html
```diff
@@ -8,6 +8,9 @@
     <div class="govuk-grid-row">
         <div class="govuk-grid-column-two-thirds">
             <h1 class="govuk-heading-l">{{ page.title }}</h1>
+            {% comment %} {% if page.is_published %}
```
marcelkornblum, line 11: I don't think this will work - `page` is _this_ page, not the target we want to link to

marcelkornblum (changes requested):

src/working_at_dit/templates/working_at_dit/working_at_dit_home.html
nicopicchio: I know but that was just a placeholder for when we know more about it.

marcelkornblum: cool - I'm not sure if we totally comment it out and commit it, or if we keep this PR open until we know more... thoughts?

nicopicchio: Not a fan of leaving commented out code if not needed but we're in the process of working on it so it might be good for everyone to have it so that we're all on the same page.

marcelkornblum: also not a fan of PRs being held open - neither is a good solution but we could timebox it to say a week or two ? Your call which is the lesser of two evils

nicopicchio: or I could convert it to draft so that the link with the jira board remains and when it's ready we can merge it?

PR #596, opened 2024-05-13
DWPF-1725 Add Navigational Link Component

alsyx:
- [ ] Tests are up to date

### DIFF
--- a/assets/stylesheets/application.scss
+++ b/assets/stylesheets/application.scss
@@ -30,6 +30,7 @@ $moj-assets-path: "/static/moj-frontend/";
 @import "components/home";
 @import "components/home_new";
 @import "components/layout";
+@import "components/navigational_link";
 @import "components/news_article";
 @import "components/news_article_comments";
 @import "components/news_category_menu";
--- a/assets/stylesheets/components/_navigational_link.scss
+++ b/assets/stylesheets/components/_navigational_link.scss
@@ -1,0 +1,11 @@
+@import url('https://fonts.googleapis.com/css2?family=Roboto+Flex:opsz,wght@8..144,100..1000&display=swap');
+
+.ws-navigational-link{
+    color: $great-blue;
+    font-family: "Roboto Flex", sans-serif;
+}
+
+.ws-navigational-link::after{
+    content: "\002192";
+    color: black;
+}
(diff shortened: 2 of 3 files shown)


src/dw_design_system/dwds/elements/navigational_link.html
```diff
@@ -0,0 +1,7 @@
+{% if self.page and self.text %}
```
marcelkornblum, line 1: we don't need any of this

this is simply an A tag with two parameters - URL and Title (unless you want to call them something else)

No logic or if/else needed

Also there's no a11y approach here - minimum would be a title attribute

src/config/settings/base.py
```diff
@@ -200,7 +200,7 @@
         "BACKEND": "django.template.backends.django.DjangoTemplates",
         "APP_DIRS": True,
         "DIRS": [
-            Path(PROJECT_ROOT_DIR) / "dw-design-system",
+            Path(PROJECT_ROOT_DIR) / "src/dw_design_system",
```
marcelkornblum, line 203:
```suggestion
            Path(PROJECT_ROOT_DIR) / "src" / "dw_design_system",
```
we definitely need the "src" do we?

src/dw_design_system/dwds/elements/navigational_link.scss
```diff
@@ -0,0 +1,26 @@
+.dwds-navigational-link{
+    display: flex;
```
marcelkornblum, line 2: we might rework this to use the every-layout pieces but this is nice

src/dw_design_system/dwds/styles/variables.scss
```diff
@@ -1,5 +1,8 @@
 :root {
+  // Colour
   --text-colour: black;
+  --dbt-blue: #00285F;
```
marcelkornblum, line 4: im going to rework these but ok for now :)

marcelkornblum (commented):
simpler, cleaner and more accessible HTML please, otherwise GTG

src/config/settings/base.py
alsyx: yes, doesn't load the templates without src @marcelkornblum

marcelkornblum (approved):

PR #599, opened 2024-05-15
Remove the ManyToMany as it causes issues with Wagtail Admin

CamLamb:
This PR fixes an issue in the Wagtail admin that cropped up due to the M2M field.

Note: The 0027 migration has already been run on staging. Therefore before this is merged, that migration should be reversed on staging with:
`./src/manage.py migrate content 0026`


marcelkornblum (approved):
argh

PR #602, opened 2024-05-15
Update the index schema to make sure it doesn't break search

CamLamb:
Running `update_index --schema-only` will update the index schema and result in search not 500'ing.
An `update_index` will still be needed if field names have changes/been added.


marcelkornblum (approved):

SamDudley (approved):
+1

PR #603, opened 2024-05-16
Hide GTM datalayer output in preview

CamLamb:
We are seeing an influx of errors because the GTM Datalayer code is being used at Wagtail page preview time. And the GTM DL code is trying to get a list of all Topic titles. However, it is failing due to the way that querysets work at preview time.

We could dig a little deeper and fix the code so that it works in this scenario, but it begs the question:
**Should we be pushing to the Datalayer when an editor is previewing a page in the Wagtail admin?**

My initial feeling is that we shouldn't (hence this PR)

### DIFF
--- a/src/core/templates/base.html
+++ b/src/core/templates/base.html
@@ -45,7 +45,9 @@
                 return false;
             };
             </script>
-            {% include "includes/gtm_datalayer_info.html" %}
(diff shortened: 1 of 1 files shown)


marcelkornblum (approved):
makes sense, I hadn't even considered preview

PR #607, opened 2024-05-17
DWPF-1716  Add CTA Card

0RWilliams:
Create new CTA (Call to Action) Card component, that requires a title with optional description.
Requires either a url or page to link to.

- [x] Tests are up to date


src/dw_design_system/dwds/stylesheet.scss
```diff
@@ -8,6 +8,7 @@
 // Components
 @forward "components/cta";
```
marcelkornblum, line 9: again with the "should we remove this" q

src/dw_design_system/dwds/components/cta_card.scss
```diff
@@ -0,0 +1,4 @@
+.dwds-cta-card{
+    background-color: #59CBE84D;
```
marcelkornblum, line 2: only CSS variables for colour, size etc please

0RWilliams: Replaced with var `--color-background-card-info`, thoughts on this name? @marcelkornblum

src/dw_design_system/dwds/components/cta_card.py
```diff
@@ -0,0 +1,45 @@
+from django.core.exceptions import ValidationError
+from wagtail import blocks
+
+
+class CTACardBlock(blocks.StructBlock):
+    """Call to action section"""
+
+    title = blocks.CharBlock(required=True, max_length=80)
+    description = blocks.CharBlock(required=False, max_length=80)
```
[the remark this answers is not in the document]
0RWilliams: With this, do we want it to be 255 char's like majority of the other titles/desc's or more limited like the old CTA (40 chars) ?

src/dw_design_system/dwds/components/cta_card.scss
marcelkornblum: I reckon one of the existing ones would be better - i made them thinking we really had two types of cards: "basic" ones and "highlighted" ones (with the dark border etc)

src/dw_design_system/dwds/components/cta_card.py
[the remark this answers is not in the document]
marcelkornblum: Just merge for now, let's discuss as a team when we have the whole page up

src/dw_design_system/dwds/components/cta_card.scss
0RWilliams: Yeah i get that, i originally was gonna use highlighted, but then thought about the figma that showed several different colours.
Actually looking at the designs now, i only see, regular, highlighted, and warning.

Will swap them out for highlighted then.

PR #609, opened 2024-05-20
DWPF-1720 Engagement card component

CamLamb:
This is an initial start at the engagement card component:
 
Screenshot 2024-05-20 at 11 13 46

- [x] Tests are up to date


src/dw_design_system/dwds/components/engagement_card.html
```diff
@@ -0,0 +1,12 @@
+{% load wagtailimages_tags %}
+
+<div class="dwds-engagement-card">
+    <a href="{{ url }}">
+        {% if thumbnail %}
+            {% image thumbnail fill-300x150 %}
+        {% endif %}
+        <h2>{{ title }}</h2>
+        <p>{{ author }}</p>
+        <time datetime="{{ date|date:'Y-m-d H:i:s' }}">{{ date }}</time>
```
marcelkornblum, line 10: is this format the GDS one? `s` seems extreme for me

marcelkornblum (approved):
minor comment

src/dw_design_system/dwds/components/engagement_card.html
CamLamb: Nahh, this is just one of the suggested ones on MDN, but it supports lots of variations, we can drop seconds for example.

src/dw_design_system/dwds/components/engagement_card.py
```diff
@@ -0,0 +1,32 @@
...
+        page = page.specific
+
+        author = page.owner.get_full_name()
+        if content_owner := getattr(page, "content_owner", None):
+            author = content_owner.full_name
+
+        context.update(
+            title=page.title,
+            author=author,
+            thumbnail=getattr(page, "preview_image", None),
```
marcelkornblum, line 27: is preview_image a wagtail field? or a common one (ContentPage) otherwise maybe?

CamLamb: Its a custom fields that is only available on NewsPage. No other image field exists AFAIK

marcelkornblum: gah. can we add one to BlogPage as well? Or is there a better solution?

CamLamb:
I have updated the styling to make this fall more in line
With the current grid setup the cards are a little narrow, but that can be addressed after this PR
Screenshot 2024-05-21 at 09 55 54

src/dw_design_system/dwds/components/engagement_card.py
CamLamb: I think the better solution would be to move the preview image up onto the content page and make it so that all content can have an image associated with it

marcelkornblum: let's do it, i see no downside

src/dw_design_system/dwds/components/engagement_card.scss
```diff
@@ -0,0 +1,62 @@
...
+    border-top: none;
+
+    &::before {
+        content: '';
+        display: block;
+        height: var(--s-1);
+        background-color: var(--color-border-dark);
+        width: calc(100% + var(--s-5) * 2);
+        margin-left: calc(var(--s-5) * -1);
+    }
```
CamLamb, line 13: This is in place to prevent angled edges on the top border since the borders are different colours

src/dw_design_system/dwds/components/engagement_card.py
CamLamb: I've made this change and added a data migration

src/dw_design_system/dwds/components/engagement_card.scss
marcelkornblum: Let's keep this and discuss a unified approach together - if this is for the "light" card we've got another 1 or 2 implementations in already, and it's making me wonder if we ought to extract it to its own subcomponent

src/content/models.py
```diff
@@ -218,6 +218,14 @@ class ContentPage(BasePage):
         blank=True, null=True, help_text="""Legacy content, pre-conversion"""
     )
 
+    preview_image = models.ForeignKey(
```
marcelkornblum, line 221: <3

marcelkornblum (approved):

src/dw_design_system/dwds/components/engagement_card.scss
CamLamb: Yeah, I thought the same, but this is for both, and is only an issue when the border on the top is different to the sides and bottom

marcelkornblum: ive got issues with the complexity, and the specific vars used, which is why i want to chat about it together with the other cards (I also dont understand the visual thing) - but IMO let's merge it so we have everything in for now then come back to it in another ticket

PR #617, opened 2024-05-23
DWPF-1783 New dwds base templates for 1 and 2 col layout

marcelkornblum:
Also made new nav page use the new base tpl

- [ ] Tests are up to date

### DIFF
--- a/src/core/templates/1_col_content_dwds.html
+++ b/src/core/templates/1_col_content_dwds.html
@@ -1,0 +1,14 @@
+{% extends "base.html" %}
+
+{% block main_content %}
+    <div class="dwds">
+        {% block pre_content %}{% endblock %}
+        <div class="ws-primary-content">
+            <article>
+                {% block content %}{% endblock %}
+            </article>
+            {% include 'includes/tag_set.html' %}
+        </div>
+        {% block post_content %}{% endblock %}
+    </div>
+{% endblock %}
--- a/src/core/templates/2_col_content_dwds.html
+++ b/src/core/templates/2_col_content_dwds.html
@@ -1,0 +1,29 @@
+{% extends "base.html" %}
+
+{% block main_content %}
+    <div class="dwds">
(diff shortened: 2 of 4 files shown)


src/core/templates/2_col_content_dwds.html
```diff
@@ -0,0 +1,29 @@
+{% extends "base.html" %}
+
+{% block main_content %}
+    <div class="dwds">
+        {% block pre_content %}
+        {% endblock pre_content %}
+        <div class="with-sidebar-right">
+            <div class="primary-content">
+                {% block pre_primary_content %}
```
alsyx, line 9: we already have `pre_content` , do we need `pre_primary_content` and `pre_secondary_content` ? If we do need them can you clarify why? @marcelkornblum

src/core/templates/1_col_content_dwds.html
```diff
@@ -0,0 +1,14 @@
...
+        {% block pre_content %}{% endblock %}
+        <div class="ws-primary-content">
+            <article>
+                {% block content %}{% endblock %}
+            </article>
+            {% include 'includes/tag_set.html' %}
+        </div>
+        {% block post_content %}{% endblock %}
+    </div>
+{% endblock %}
```
alsyx, line 14: is this template for another / sub type of navigation page?

src/core/templates/2_col_content_dwds.html
marcelkornblum: it's more "maybe" than "yes" - but my e.g. was in the new firebreak HP we moves the site banner to a spot that would match this

src/core/templates/1_col_content_dwds.html
marcelkornblum: removed now :)

nicopicchio (approved):

alsyx (approved):

PR #619, opened 2024-05-24
Fix new homepage news styling

CamLamb:
Prevent css overspill


CamLamb:
Including the new design system styles on all pages caused this issue as it adds some global styles to classes that already existed (specifically the `frame` class)

I have added a workaround for now of ensuring that the layout frame css only applies inside elements with the `dwds` class

nicopicchio (approved):

src/dw_design_system/dwds/styles/layout-primitives.scss
```diff
@@ -1,256 +1,259 @@
...
-.stack {
+  .stack {
     display: flex;
     flex-direction: column;
     justify-content: flex-start;
-}
+  }
 
-.stack > * {
+  .stack>* {
```
marcelkornblum, line 11: I don't like all this reformatting - IMO the way your linter likes it is much less readable...

marcelkornblum (commented):
I'm not sure I actually like these changes TBH - and the fact they're in components we don't even actually use (and essentially have for reference) makes me like them even less.

I wonder if we ought yo just remove these unused ones altogether and just make sure we implement them more carefully when we need them

PR #620, opened 2024-05-24
DWPF-1763 Navigation card title made optional

nicopicchio:
The navigation card title is now optional. When the title is not provided on the nav card component then the linked page title will be rendered instead.
Also added help text to explain the behaviour to users.


src/dw_design_system/dwds/components/navigation_card.py
```diff
@@ -4,8 +4,12 @@
 class NavigationCardBlock(blocks.StructBlock):
     """A nav card to direct users"""
 
-    title = blocks.CharBlock(required=True, max_length=30)
-    page = blocks.PageChooserBlock(required=True)
+    page = blocks.PageChooserBlock()
+    title = blocks.CharBlock(
+        required=False,
+        max_length=30,
+        help_text="By default, the navigation card component will render the name of the page you linked above. If you have to change this behavior, add an ad hoc title in the box below. Please use sparingly.",
```
marcelkornblum, line 11:
```suggestion
        help_text="By default, the name of the page will be the title. Override it using the below field.",
```

marcelkornblum (approved):
Made a version of the help text that was a bit snappier (and without americanised spelling :) )

PR #621, opened 2024-05-28
DWPF-1764 Highlight Engagement Cards with News and BlogPosts

alsyx:
News and Blog Post engagement cards to be highlighted in darker (blue) card colour.
- [x] Tests are up to date


nicopicchio (approved):

src/dw_design_system/dwds/components/engagement_card.py
```diff
@@ -38,7 +39,10 @@ def get_context(self, value, parent_context=None):
             thumbnail=getattr(page, "preview_image", None),
             date=page.published_date,
             url=page.url,
-            is_news=isinstance(page, NewsPage),
+            is_news_or_blog_post=[
```
marcelkornblum, line 42: to me this is the wrong level of abstraction. I would prefer this var to be something like `is_highlighted` instead

src/dw_design_system/dwds/components/engagement_card.py
```diff
@@ -4,14 +4,15 @@
 
 
 class EngagementCardBlock(blocks.StructBlock):
-    page = blocks.PageChooserBlock(page_type="content.ContentPage")
+    page = blocks.PageChooserBlock(page_type=["content.BlogPost", "news.NewsPage"])
```
marcelkornblum, line 7: i tihnk remove restrictions from this, we have no idea if this is right or not. let it be anything at this stage

CamLamb: It came from a discussion with Paul as he was saying that only "engaging content" should be able to be selected for use in this component

I guess it wouldn't really make sense for someone to be able to link to the privacy policy or something with a highlighted card

marcelkornblum: right but we're also so early in the process that artificially adding restrictions feels like it'll cause us more work. 

we already control who can use this component; if it turns out that we want it to only be those types then we don't need the exposed api about light  and dark styling, or maybe we want to leave that open but we decide the easiest way for editors to use it is with two different python block types. 

Thing is we are too early to know either whether the restriction is needed or how best to implement it, so our heuristic should be to leave it open, knowing we're covering it by having limited exposure to it until we know more.

PR #645, opened 2024-06-12
Improve index update

CamLamb:
This PR aims to improve the search index update logic to reduce the risk of downtime.
Search downtime will still happen with these changes, but the goal is to reduce the likelihood of it happening, and when it does happen, we inform the user more gracefully.

## The current issues
Running a schema only update while also using an atomic rebuild process results in an empty index being pushed. The Wagtail documentation doesn't warn users of this behaviour, but it appears that this is what is happening.
Not running a schema only update results in our search views returning 500 errors as the mapping of the documents doesn't match the expected mapping of our code, resulting in the query failing.

## Proposed solution
We update the logic so that we no longer do a schema only update, this will result in 500 errors.
We catch the error and raise a nicer more user friendly message that says something along the lines of "Search is currently down, it should be back up in X minutes" etc.
We reduce the likelihood of users seeing this on deployment, by only running the update index command on deployments that we think NEED the index updating.

The outcome of this will be that deployments that change search mappings will result in an update index and users seeing an error message (a much nicer error message). However, most deployments shouldn't effect the search at all.

## Changes made so far:
 - [x] No more schema only updates
 - [ ] Catch the search error and return a nicer error message to the user
 - [x] Only update search on deployments that alter the mapping

### DIFF
--- a/src/core/management/commands/update_index_async.py
+++ b/src/core/management/commands/update_index_async.py
@@ -1,4 +1,6 @@
+from django.core.cache import cache
 from django.core.management.base import BaseCommand
+from wagtail.search.management.commands.update_index import Command
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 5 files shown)


src/core/management/commands/update_index_async.py
```diff
@@ -7,8 +9,41 @@ class Command(BaseCommand):
...
+            # Get the search mapping dict
+            sm_dict = get_indexed_mapping_dict()
+            # Get the hash of the current search mapping dict
+            new_sm_hash = hash(frozenset(sm_dict.items()))
+
+            # If the hash of the current search mapping dict is different from the
+            # hash of the search mapping dict that was stored in the cache, then
+            # update the search index.
+            if sm_hash != new_sm_hash:
+                cache.set("search_mapping_hash", new_sm_hash)
```
CamLamb, line 42: I'm not currently settled on using the cache for this, but this is just to capture the concept first.
Happy to discuss alternatives as the negative impact of this not being in the cache is higher than "just a longer page load time"

CamLamb:
Fixed with latest changes to search indexing and query caching

PR #660, opened 2024-07-16
Pattern Lib 2: Restructure the components

CamLamb:


CamLamb:
Closing this after discussions as it isn't quite what we need at the moment

PR #663, opened 2024-07-16
Add layouts to the DWDS

CamLamb:


marcelkornblum:
I've got quite a lot of feedback about this, and wouldn't mind a discussion.

Top of the list though, is I don't think we will be sticking with anything 1 col most likely, and I think our layouts will be becoming more diverse. I don't think naming 1_col and 2_col is forward thinking in that context.

Also, I'm not totally convinced we need or want the layouts inside the DS, as opposed to in `core`. Maybe they should be but I'm just not convinced. It's kind of related to the above comment in a way.

Either way please hold off this until it's been chatted about with the team :)

CamLamb:
OLD AND UNUSED

PR #664, opened 2024-07-16
A simple Components page

CamLamb:
dwds-components

### DIFF
--- a/src/dw_design_system/templates/dw_design_system/components.html
+++ b/src/dw_design_system/templates/dw_design_system/components.html
@@ -1,0 +1,12 @@
+{% extends "dwds_content.html" %}
+{% load static %}
+{% load wagtailcore_tags %}
+{% load wagtailimages_tags %}
+{% load site_alert %}
+{% load dw_design_system %}
+
+{% block primary_content %}
+    {% for component_template, component_context in components %}
+        {% render_component component_template component_context %}
+    {% endfor %}
+{% endblock primary_content %}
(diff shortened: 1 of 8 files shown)


src/dw_design_system/templates/dw_design_system/components.html
```diff
@@ -0,0 +1,71 @@
...
+{% endblock page_title %}
+
+{% block primary_content %}
+    {% for component in components %}
+        {% include "dw_design_system/partials/component.html" %}
+    {% endfor %}
+{% endblock primary_content %}
+
+{% block scripts %}
+    <style type="text/css">
```
marcelkornblum, line 19: ought this to be a css file?

src/dw_design_system/utils.py
```diff
@@ -0,0 +1,114 @@
...
+from wagtail.images.models import Image
+
+
+DATETIME_STR = "datetime"
+IMAGE_STR = "Image"
+
+
+def get_components():
+    thumbnail_file = Image.objects.first()
+    return [
```
marcelkornblum, line 14: not a biggie but for me this being a data file would be preferable; easier to maybe run a scan and check it has the right number of elements in it or whatever

marcelkornblum (approved):
this is very smart, i like it :)

src/dw_design_system/templates/dw_design_system/components.html
CamLamb: Yeah, we can move it to a separate CSS file in a separate PR. I'm going to merge this for now so I can unblock people 😄

src/dw_design_system/utils.py
CamLamb: Added a todo :)

PR #674, opened 2024-07-23
INTR-189 CTA Card updates

CamLamb:
- [x] Tests are up to date


package.json
```diff
@@ -3,6 +3,7 @@
     "author": "Department for Business and Trade",
     "private": true,
     "dependencies": {
+        "@fortawesome/fontawesome-free": "^6.6.0",
```
marcelkornblum, line 6: aah - I was hoping we could avoid these sorts of libs - and if we need them ideally it wouldn't be via scss since it makes it harder to move to pure CSS later. Not that I understand allthe thinking going on mind, just a pref

src/dw_design_system/dwds/elements/navigational_link.html
```diff
@@ -1,3 +1,4 @@
 <a href="{{ url }}" class="dwds-navigational-link">
     <span>{{ text }}</span>
+    <i class="fa-solid fa-arrow-right"></i>
```
marcelkornblum, line 3: are there a11y considerations with an empty inline tag, and an `i` one at that?

webpack.config.js
```diff
@@ -32,7 +32,7 @@ module.exports = {
 
       // Copy all images and fonts to the output directory ignoring moj assets
       {
-        test: /\.(png|jpe?g|gif|woff2?|svg|ico)$/i,
+        test: /\.(png|jpe?g|gif|woff2?|ttf?|svg|ico)$/i,
```
marcelkornblum, line 35: is this for FA?

CamLamb: Yup

src/dw_design_system/dwds/elements/navigational_link.html
CamLamb: https://docs.fontawesome.com/web/dig-deeper/accessibility#font-awesome-icons-and-accessibility
IMO it shouldn't be an issue as this scenario seems like a decorative one, we could add `aria-hidden="true"` to ensure it doesn't show to screen readers.

package.json
CamLamb: Do we use fontawesome (or a similar lib)
- If we do, should we load it in with JS instead (As per [the docs](https://docs.fontawesome.com/web/dig-deeper/accessibility#when-is-auto-accessibility-available) this will add [Auto Accessibility](https://docs.fontawesome.com/web/dig-deeper/accessibility#making-icons-accessible-with-auto-accessibility))
- If we don't, we will need to discuss with the Designers the options to ensure we are all working from the same icons

I was speaking with Cheryl earlier and she mentioned that some of the Icons might be coming from Great, so it'd be good to work out if/how we use those.

src/dw_design_system/dwds/elements/navigational_link.html
marcelkornblum: to me it feels like layers of workarounds creeping in - maybe inevitable but it's painful to see the purity diminishing. We couldn't just use characters right?

webpack.config.js
CamLamb: Resolving this as FA has been removed

src/dw_design_system/dwds/elements/navigational_link.html
CamLamb: Resolving this as FA has been removed

package.json
CamLamb: Resolving this as FA has been removed

nicopicchio (approved):

PR #705, opened 2024-08-21
Fix the search results to include the excerpt

CamLamb:
Return the specific page models so we have access to `.excerpt`


marcelkornblum (approved):
thank you!

PR #709, opened 2024-08-27
INTR-248 Add the Event page template

alsyx:
On the event page `information panel` I used normal buttons as CTAs. When we have a component for CTA button or "button with an arrow", these can be updated.

- [ ] Tests are up to date


src/core/templates/dwds_content.html
```diff
@@ -15,14 +15,14 @@
                 {% endblock pre_content %}
                 {% block page_title %}
                     <h1>{{ page.title }}</h1>
-                    {% if page.description %}<p>{{ page.description }}</p>{% endif %}
+                    {% if page.description %}<p style="font-size: var(--text-xlarge);">{{ page.description }}</p>{% endif %}
```
CamLamb, line 18: Can we change this from an inline style to a class with the style please?

src/dw_design_system/dwds/elements/information_panel.scss
```diff
@@ -0,0 +1,22 @@
+.dwds-card.dwds-information-panel {
+    --card-bottom-border-height: 0px;
```
CamLamb, line 2: There is a new approach, if you pull in the latest changes from main, instead of overriding the variable, you can just add the `card-without-shadow` class to the component.

src/events/templates/events/event_page.html
```diff
@@ -6,41 +6,88 @@
...
-        <p>Room: {{ page.room|default_if_none:"-" }} Capacity: {{ page.room_capacity|default_if_none:"-" }}</p>
+            <div>
+                <p>Audience</p>
+                <h4>{{ page.get_audience_display|default_if_none:"-" }}</h4>
+            </div>
+            <div>
+                <p>Event type</p>
+                {% if is_online or is_in_person %}
+                    <h4>{{ page.get_event_type_display }}</h4>
+                {% elif is_hybrid %}
```
CamLamb, line 61: One to think about, I'm not too fussed either way:
this could be achieved with `elif is_online and is_in_person`

I can see pros/cons to both approaches so I'll let you make a call on which feels best

src/events/templates/events/event_page.html
```diff
@@ -6,41 +6,88 @@
...
+                </div>
+                {% if page.location %}
+                    <div>
+                        <p>Room</p>
+                        <h4>
+                            {% if page.room %}
+                                {{ page.room }}
+                                {% if page.room_capacity %}with {{ page.room_capacity }} seats{% endif %}
+                            {% else %}
+                                TBC
```
CamLamb, line 34: TBC only makes sense if an in person event will definitely have a room at some point.
I'm ok with this approach if we are confident that all in person events will eventually have a room assigned to them

src/events/templates/events/event_page.html
```diff
@@ -6,41 +6,88 @@
...
+                        </h4>
+                    </div>
+                    {% if page.room %}
+                        <div>
+                            <p>Capacity</p>
+                            <h4>
+                                {% if page.room_capacity %}
+                                    {{ page.room_capacity }} seats
+                                {% else %}
+                                    -
```
CamLamb, line 46: Instead of having an empty capacity, if the `room_capacity` field is empty, let's just hide this section

src/events/templates/events/event_page.html
```diff
@@ -6,41 +6,88 @@
...
-            <a href="{{ page.event_recording_url }}"
-               class="dwds-button dwds-button--inline dwds-button--secondary">View event recording</a>
-        {% endif %}
-
-        <p>Audience: {{ page.get_audience_display }}</p>
-        <p>Location: {{ page.location|default_if_none:"-" }} {{ page.location.city|default_if_none:"" }}</p>
-        <p>Room: {{ page.room|default_if_none:"-" }} Capacity: {{ page.room_capacity|default_if_none:"-" }}</p>
+            <div>
+                <p>Audience</p>
+                <h4>{{ page.get_audience_display|default_if_none:"-" }}</h4>
```
CamLamb, line 55: Similarily, if there is no audience to display, let's just hide the section instead of showing a blank line

src/dw_design_system/dwds/styles/layout-primitives.scss
```diff
@@ -43,7 +43,7 @@
   // Sidebar
   // https://every-layout.dev/layouts/sidebar/
   .with-sidebar,
-  .with-sidebar-right {
+  .with-content-right {
```
CamLamb, line 46: As discussed in slack, let's continue with the old name and just keep using this class as there is no need to define a new class just because of the name (when all of the styles will be identical)

src/events/templates/events/event_page.html
alsyx: This was from my discussions with Cheryl, I think most of the in person events will have a room but not sure if all of them will have it. How can we clarify this?

src/dw_design_system/dwds/elements/information_panel.scss
alsyx: yes, but i wanted `dwds-information-panel` class not to have that shadow whenever/wherever we use it. If i add that class in html we always have to add it. If that's our team's approach then i can update it of course.

src/events/templates/events/event_page.html
alsyx: I think i will use `else` instead, i don't know why i used `elif`:
```suggestion
               {% if is_online or is_in_person %}
                    <h4>{{ page.get_event_type_display }}</h4>
                {% else %}
```

src/events/templates/events/event_page.html
CamLamb: I guess we won't know for sure until people are using it. Maybe we raise this when we are next discussing it as a wider team?

src/dw_design_system/dwds/styles/temporary-base.scss
```diff
@@ -87,6 +87,10 @@ body {
         border-left: 0;
         color: var(--color-hr);
     }
+
+    .p-with-xlarge-text {
```
CamLamb, line 91: I think this class should be just `text-xlarge` and should be in the typography stylesheet. That way we can use it on other elements as there is nothing specific to this style for a `<p>` tag

CamLamb (approved):
Just one comment to address

CamLamb (approved):

CamLamb (approved):

PR #711, opened 2024-08-29
INTR-247 Show events on new homepage

nicopicchio:
- [ ] Tests are up to date


src/dw_design_system/dwds/components/banner_card.scss
```diff
@@ -1,23 +1,11 @@
 .dwds-card.dwds-banner-card {
     h3 {
-        color: var(--color-text-highlight);
```
marcelkornblum, line 3: why are we removing these?

src/dw_design_system/dwds/components/cta_card.html
```diff
@@ -1,6 +1,6 @@
 {% extends "dwds/elements/extendable/card.html" %}
 
-{% block card_container_classes %}dwds-cta-card dwds-card-blue{% endblock %}
+{% block card_container_classes %}dwds-cta-card card-without-shadow dwds-card-blue{% endblock %}
```
marcelkornblum, line 3: i dont like the way we have inconsistent naming for all of these. I would choose `dwds-cta-card without-shadow alt-colour` but if we're sticking `dwds-` on everything we should do that and if not we should remove it. it's a bigger discussion but in any case we ought not to be making things less consistent.

src/dw_design_system/dwds/components/cta_card.scss
```diff
@@ -1,14 +1 @@
...
-        color: var(--color-text-highlight);
-        text-decoration-color: var(--color-text-highlight);
-
-        p {
-            color: var(--color-text-highlight);
-        }
-    }
-}
\ No newline at end of file
+.dwds-card.dwds-cta-card {}
```
marcelkornblum, line 1: if you're taking the declaration out of the codebase clean it up and remove the class altogether (inc where it's referenced) please

src/dw_design_system/dwds/components/engagement_card.html
```diff
@@ -22,16 +22,23 @@
...
     {% if excerpt %}<p>{{ excerpt|truncatewords:16 }}</p>{% endif %}
 {% endblock card_content_body %}
 
 {% block card_footer_body %}
-
-    <time>
-        Added {{ date|naturalday:'j F Y' }}
-        {% if date|naturalday in 'yesterday, today' %}at {{ date|time:'g:iA'|lower }}{% endif %}
-    </time>
+    {% if date %}
```
marcelkornblum, line 36: is `date` specific / futureproof enough for this component?

src/dw_design_system/dwds/elements/extendable/card.scss
```diff
@@ -14,21 +18,43 @@
...
         margin-bottom: var(--s-5);
         margin-top: var(--s-5);
     }
+    
+    h2,
+    h3 {
+        color: var(--card-body-heading-color);
+    }
 
     &.dwds-card-blue {
```
marcelkornblum, line 46: i would really prefer this name to be less about the implementation and more about the purpose - rather than "blue" could it be "alternative" or something that makes more sense (ie even if we change the colour)?

src/dw_design_system/dwds/elements/extendable/card.scss
```diff
@@ -14,21 +18,43 @@
...
         --card-body-bg-color: var(--card-bg-blue);
+        --card-body-heading-color: var(--color-text-highlight);
+        --card-body-link-color: var(--color-text-highlight);
```
marcelkornblum, line 49: are we sure we're making text and links the same colour? that's going to break a11y isnt it?

src/events/utils.py
```diff
@@ -0,0 +1,27 @@
+from datetime import time
+
+from events.models import EventPage
+
+
+def format_time(time_obj: time) -> str:
```
marcelkornblum, line 6: is there not a built-in djangoish way to do this? and ought we not to be making this sort of thing more global to the site?

src/home/templates/home/home_page_new.html
```diff
@@ -56,7 +55,10 @@ <h1>Intranet</h1>
                 </div>
 
                 {% pages_to_cards news_items as news_cards %}
-                {% include "dwds/components/card_group.html" with card_group_title="Latest DBT news" cards=news_cards footer_link="/news-and-views/" footer_link_text="View all news" %}
+                {% include "dwds/components/card_group.html" with card_group_title="Latest DBT news" cards=news_cards footer_link="/news-and-views/" footer_link_text="View all news" show_hr=True %}
+
+                {% pages_to_cards events as event_cards %}
+                {% include "dwds/components/card_group.html" with card_group_title="Events" cards=event_cards grid=True blue_bg=True %}
```
marcelkornblum, line 61: i feel like the grid being baked in really breaks the model (at least in my head) of our components owning their responsiblities - it feels like the grid ought to be _inside_ the card or we risk the card getting so bloated its unwieldy to look after. I might be in the wrong about this though.

marcelkornblum (commented):
I have many comments from small to big. 

I might be wrong on a few of them but wouldn't mind a chat to go through if poss....

src/dw_design_system/dwds/components/banner_card.scss
CamLamb: We're moving the logic to turn the text to white to be effected by the css class that changes the background to blue/red

src/dw_design_system/dwds/components/cta_card.html
CamLamb: Yah, I have been thinking that coming up with some guidance on this would be a good idea as I find myself "just going with something for now" and it ends up sticking

src/home/templates/home/home_page_new.html
CamLamb: This could do with a catchup 😄

src/dw_design_system/dwds/components/cta_card.html
CamLamb: Discussed with Nico and we will leave this as is for now as I am writing up a refactor ticket where we can agree on a consistent approach and bring everything inline together at the same time.

src/dw_design_system/dwds/components/engagement_card.html
CamLamb: Will address this as part of a refactor ticket, likely renaming to "created_date" (for more specificity)

src/dw_design_system/dwds/components/cta_card.scss
CamLamb: @nicopicchio delete the file and remove references as per this comment

src/dw_design_system/dwds/elements/extendable/card.scss
CamLamb: I will add a ticket to ensure we discuss this as a team and with Design

src/dw_design_system/dwds/elements/extendable/card.scss
CamLamb: I'm adding a ticket to discuss this with design and address it separately to this work.

src/events/utils.py
CamLamb: Let's move `format_time` into `core.utils`

src/home/templates/home/home_page_new.html
CamLamb: We've spoken about this and we have some theoretical/conceptual ideas to discuss as a team. We can address this separately to this work.

assets/stylesheets/components/_home_new.scss
```diff
@@ -73,6 +73,15 @@
...
+
+  .post-title-container {
+    flex-direction: column;
+    gap: var(--s-5)
+  }
+
+  .post-title {
+    margin: 0;
+  }
 }
```
CamLamb, line 85: It looks like the only place these classes are used are in the `engagement card` let's move them into the engagement card stylesheet 😄

nicopicchio: moved

CamLamb (approved):

PR #712, opened 2024-08-29
INTR-272 Update admin to allow managing above fold content on home page

CamLamb:
This PR adds the option to reference 3 pages on the home page admin, these are used on the new homepage layout to show content in the "above fold" cards.

We also update the comments model so that it can be used on all content pages, still leaving the logic so that it's only available on news pages. This was done so that we have the ability to get comment counts in ContentPage queries. Also, we will likely be adding comments to other pages in the future.

**Data migrations:**
- All comments should still retain the data on them
- The new "priority pages" will be populated with the first 3 HomeNewsOrder objects

- [x] Tests are up to date


src/home/models.py
```diff
@@ -131,10 +159,17 @@ def clean(self):
...
     subpage_types = []
-
     promote_panels = []
+    content_panels = [
+        InlinePanel(
+            "priority_pages",
+            label="Priority page",
+            heading="Priority pages",
+            min_num=0,
+            max_num=3,
```
marcelkornblum, line 170: do we want to encode this here when it'll soon be flexible; would changing it trigger a migration?

marcelkornblum (approved):
feels like 2 PRs but looks good :)

src/home/models.py
CamLamb: This lets it be a deployable state, removing this makes the admin a lil funky if the layout stuff is still being discussed

PR #714, opened 2024-09-02
Fix - Added missing template files for blogs to work/use new styling

0RWilliams:
Quick little fix to resolve missing template files for Blogs.


CamLamb:
Are blogs currently accessible?
If not then I think we should take the same approach as the Navigation Page, where they are only visible in the new styles.
With my goal here being that we don't want to maintain 2 templates if no one has ever seen a blog in the old design

0RWilliams:
> Are blogs currently accessible? If not then I think we should take the same approach as the Navigation Page, where they are only visible in the new styles. With my goal here being that we don't want to maintain 2 templates if no one has ever seen a blog in the old design

Blogs are not currently accessible in production. Fair enough, updated to use base template.

PR #722, opened 2024-09-09
INTR-304 Priority pages uniqueness validation

nicopicchio:
- [ ] Tests are up to date


src/home/validators.py
```diff
@@ -3,10 +3,15 @@
...
 
     p = Page.objects.get(id=value).specific
     if not issubclass(p.__class__, HOME_PRIORITY_PAGE_TYPES):
         raise ValidationError(
             "Only NewsPage objects can be added to the priority pages list",
         )
+
+    if HomePage.objects.filter(priority_pages__page_id=value).exists():
+        raise ValidationError(
+            "This page is already in the priority pages list.",
```
CamLamb, line 16: This sentence is good, but it might be more useful with an action after
"Please remove one of the references to 'Page title'"

nicopicchio: Should I also change the one above that checks if it's a `NewsPage`? I don't think the user will understand "NewsPage object". Maybe something along the lines of "Only news and event pages can be added to the priority pages list?

nicopicchio: Actually, instead of "Please remove one of the references to 'Page title'" it could refer to the the action of having to choose another page, which is the option you get from the menu.
image
Maybe "Please select another page"?

PR #738, opened 2024-09-16
INTR 322 - Enable CSP logging into Sentry

CaitBarnard:
Add CSP headers and log any violations directly to sentry.

- Currently set to report only
- New env variable for the sentry report endpoint, if empty then nothing happens
- Any violations will appear in the console 

- [ ] Tests are up to date


src/core/views.py
```diff
@@ -102,6 +104,33 @@ def page_problem_found(request):
     )
 
 
+@csrf_exempt
```
CaitBarnard, line 107: I would appreciate a second opinion on whether it makes sense to exempt this view from the token or if there's a reasonable way to add this given it's a basic view with no .html

src/config/settings/base.py
```diff
@@ -727,3 +728,16 @@ def filter_transactions(event, hint):
...
+# Content Security Policy header settings
+CSP_DEFAULT_SRC = "*"  # Allows all sources
+CSP_SCRIPT_SRC = "*"
+CSP_SCRIPT_SRC_ATTR = "*"
+CSP_SCRIPT_SRC_ELEM = "*"
+CSP_IMG_SRC = ("*", "data:")
+CSP_MEDIA_SRC = "*"
+CSP_FRAME_SRC = "*"
+CSP_FONT_SRC = "*"
+CSP_CONNECT_SRC = "*"
```
marcelkornblum, line 741: have we tested that these permissive settings will report "breaches" (I guess I mean all things that CSP _could_ apply to)? I feel like that aspect isn't yet well understood in the team (or at least not by me!)...

CaitBarnard: I've tested a few of them, hence why the CSP IMG SRC is different as the wildcard strangely didn't cover a data element that was triggering a violation. Will locking them down be part of a separate ticket?

CamLamb (approved):

PR #744, opened 2024-09-18
Header tweaks

CamLamb:
An interim site header while we wait for the IA work so that we can design/build a better one in the DWDS

Screenshot 2024-09-18 at 15 12 44
Screenshot 2024-09-18 at 15 13 10


marcelkornblum (approved):
inline styles FTW

PR #746, opened 2024-09-19
INTR 323 New pagination component in DWDS

0RWilliams:
Adds 2 new components inside of DWDS, 
-`dwds/pagination`
-`dwds/link_navigate`

pagination vars:
`pages` - generalised list of pagination supported documents.

link_navigate vars:
`previous_text` - previous navigation text, defaults to 'Previous'.
`previous_url` - url to previous to when clicking on previous_text.
`next_text` - next navigation text, defaults to 'Next'.
`next_url` - url to navigate to when clicking on next_text.

Numerical navigation:
image
image
image

Previous/Forward navigation:
image
image
image

- [x] Tests are up to date


CamLamb:
> A couple of general comments:
> 
> * do we want to expose so many variables? What's the case for overriding the text? Personally I would rather hide that level of stuff so we are enforcing consistency; remember this DS is aimed solely at the intranet.
> * On the other hand we're not for example exposing the name of the GET var we use to drive pagination, it's hard-coded as `page`. Which is OK but it also looks like some of your variables are expecting to be associated with a Django pagination object.
> 
> Basically i tihnk the API is inconsistent; my preference would be to explicitly require passing in an instantiated Django Paginator and then referencing that to get hold of all the URLs and things (which you do for some links now but not direclty and not for all).

Not all scenarios are covered by a Django paginator object. For example, in the events listing page, we want to move forward or backward a month, this is really a filter change not a pagination. The django Paginator class takes a list of objects and a page count and paginates based on that, not based on a filter.

We could potentially write our own custom Paginator class that forces the events filtering to work like pagination, but that feels a little overkill.
Alternatively, we could just use 2 "links" and leave the scope of Paginator to just scenarios where a Paginator class can be used

src/news/templates/news/news_home_new.html
```diff
@@ -51,11 +51,12 @@ <h2 class="govuk-heading-m">
...
         {# Only show pagination if there is more than one page to click through #}
         {% if posts.paginator.num_pages > 1 %}
-            <div class="container">
+            {% include "dwds/components/pagination.html" with pages=posts %}
+            {% comment %} <div class="container">
                 <div class="row">
                     <div class="col-lg-12">{% include "dwds/components/pagination.html" with pages=posts %}</div>
                 </div>
-            </div>
+            </div> {% endcomment %}
```
CamLamb, line 59: This commented code can be removed :)

src/dw_design_system/dwds/components/link_navigate.scss
```diff
@@ -0,0 +1,22 @@
+.dwds-link-navigate{
+    display: flex;
+    text-decoration: none;
```
CamLamb, line 3: is this needed? it looks like the links should be underlined

CamLamb (approved):

PR #749, opened 2024-09-20
"View all events" link flag

CamLamb:
Put the view all events link behind a flag as we don't want it to show at the moment


SamDudley:
We should document current feature flags.

SamDudley (approved):

PR #752, opened 2024-09-23
INTR-339 Update URLFields to have max_limit of 2048 chars

alsyx:
- Django's URLField is a CharField therefore it has max_limit = 200 characters. 
- Created a custom URLField with 2048 character limit, which is the maximum length of a URL.

- [ ] Tests are up to date


CamLamb:
Do we really have to change the whole underlying field?
Does setting a `max_length` value to something large not work?

CamLamb (approved):
Approving on the condition that you have checked the data migrates fine and you can save a long URL in the new field

CamLamb (approved):
Migration looks good!

PR #753, opened 2024-09-23
INTR-318 Event listing page updates and filtering

CamLamb:
- [x] Tests are up to date

- Event listing page at `/events/` that shows the events for the current month
- Ability to navigate forwards/backwards in months to view events for future/past months/years
  - `/events/{YEAR}/` redirects to `/events/{YEAR}/1/`
  - `/events/{YEAR}/{MONTH}/` shows events for the given month/year
- All pages have the filters:
  - Event format (in person/online)
  - Location (city)


david-okeke1337 (commented):
Looks good, however I noticed it loops back around to the start/end of the year if you reach month 1/12. Is that intended? Is the only way to change the year via the URL?

CamLamb:
> Looks good, however I noticed it loops back around to the start/end of the year if you reach month 1/12. Is that intended? Is the only way to change the year via the URL?

Good catch! I've just pushed some changes that should fix this 😄

david-okeke1337 (approved):

PR #754, opened 2024-09-24
INTR 317 - Clean up home page feature flags

CaitBarnard:
Remove code and conditionals related to the new homepage feature flag
- Remove .html files for old homepage
- Add error handling for empty priority pages

---

- [ ] Tests are up to date


src/home/models.py
```diff
@@ -302,45 +289,27 @@ def get_context(self, request, *args, **kwargs):
...
-        if not is_new_homepage:
-            context["how_do_i_items"] = (
-                HowDoI.objects.filter(include_link_on_homepage=True)
-                .live()
-                .public()
-                .order_by(
-                    "title",
-                )[:10]
-            )
+        whats_popular_items = WhatsPopular.objects.all()
```
CaitBarnard, line 307: This code was previously commented. Is it needed?

src/core/templates/includes/header_new.html
```diff
@@ -51,12 +46,6 @@
         </div>
```
CaitBarnard, line 46: The border styling on the search box and styling on the search button have vanished

src/news/models.py
```diff
@@ -209,8 +207,7 @@ def serve(self, request, *args, **kwargs):
         context = self.get_context(request, **kwargs)
         context["comment_form"] = CommentForm()
 
-        if flag_is_active(request, FEATURE_HOMEPAGE):
-            self.template = self.template.replace(".html", "_new.html")
+        self.template = self.template.replace(".html", "_new.html")
```
CamLamb, line 210: Let's drop this replace logic completely and rename the templates to no longer need the `_new` suffix.

src/home/templates/home/home_page_new.html
```diff
@@ -17,8 +17,6 @@
 {% endblock body_classes %}
 
 {% block pre_content %}
-    {% include "home/includes/home_banner.html" with hide_hr=True %}
-
     {% comment %}
```
CamLamb, line 20: We can strip out this commented code and bring it back if needed at a later date.

src/home/models.py
CamLamb: Please remove it 😄 
I also thing there is a reference to it in some commented out code in the new home page template, it can be removed from there also

src/home/models.py
```diff
@@ -302,45 +290,27 @@ def get_context(self, request, *args, **kwargs):
...
         govuk_feed = cache.get("homepage_govuk_news")
-        if is_new_homepage:
-            govuk_feed = [
-                {"url": obj.links[0].href, "text": obj.title.value}
-                for obj in govuk_feed
-            ]
+        govuk_feed = [
+            {"url": obj.links[0].href, "text": obj.title.value} for obj in govuk_feed
+        ]
         context["govuk_feed"] = govuk_feed
```
CamLamb, line 296: Where possible, we can remove variables that are only used to hold a place and don't provide any other value.

It can provide value by making the code a little cleaner and easier to read, but in most scenarios this is not the case.

```suggestion
        context["govuk_feed"] = [
            {"url": obj.links[0].href, "text": obj.title.value} for obj in govuk_feed
        ]
```

src/home/models.py
```diff
@@ -234,53 +231,44 @@ def to_page_counts(self) -> list[int]:
...
+            "-first_published_at",
+        )
+
+        context.update(
+            priority_pages=priority_pages,
+            events=EventPage.objects.live()
+            .public()
+            .filter(event_date__gte=timezone.now().date())
+            .exclude(id__in=priority_page_ids)
+            .order_by("event_date", "start_time")[:6],
```
CamLamb, line 268: While we are here, let's move the EventPage queryset out of the context update and assign it to a variable like we do with `news_items` (this is one of those scenarios that makes the code a little cleaner)

src/core/templates/includes/header_new.html
CamLamb: Good catch!
That's because the CSS is targeting the `data-flag-new-homepage` we can now drop this target and override the old styling to be the new styles

src/core/templates/base.html
```diff
@@ -84,11 +84,7 @@
         <a href="#main-content"
            data-module="govuk-skip-link"
            class="govuk-skip-link">Skip to main content</a>
-        {% if FEATURE_FLAGS.new_homepage %}
-            {% include "includes/header_new.html" %}
-        {% else %}
-            {% include "includes/header.html" %}
-        {% endif %}
+        {% include "includes/header_new.html" %}
```
CamLamb, line 87: Since we've deleted the old `header.html` we can drop the `_new` suffix from this file name.

src/content/models.py
```diff
@@ -509,10 +505,8 @@ def get_context(self, request, *args, **kwargs):
         context["tag_set"] = tag_set
 
         # override page base when the new homepage is active.
-        is_new_homepage = flag_is_active(request, FEATURE_HOMEPAGE)
-        if is_new_homepage:
-            context["override_base"] = "dwds_content.html"
-            context["override_content"] = "primary_content"
+        context["override_base"] = "dwds_content.html"
+        context["override_content"] = "primary_content"
```
CamLamb, line 509: It doesn't look like these are actually used anywhere so these 2 lines can go 😄

src/content/models.py
```diff
@@ -181,16 +179,14 @@ def published_date(self):
...
         return self.template
 
     def serve(self, request):
         response = super().serve(request)
 
-        if flag_is_active(request, FEATURE_HOMEPAGE):
-            self.template = self.get_template(request)
+        self.template = self.get_template(request)
 
         return response
```
CamLamb, line 191: I think that this whole serve method can be removed now

src/content/models.py
```diff
@@ -181,16 +179,14 @@ def published_date(self):
         return self.last_published_at
 
     def get_template(self, request, *args, **kwargs):
-        if flag_is_active(request, FEATURE_HOMEPAGE):
-            self.template = self.template.replace(".html", "_new.html")
+        self.template = self.template.replace(".html", "_new.html")
```
CamLamb, line 182: This get template method can be removed and all the non suffixed templates replaced by the suffixed ones

CamLamb (commented):
Great work!
I've added some comments for your next pass

src/home/models.py
CaitBarnard: Are you suggesting we inline the variable? These tripped me up before as they look like they're being reassigned  immediately but it's being mutated instead as the line below uses it in the for loop.

`govuk_feed = [
            {"url": obj.links[0].href, "text": obj.title.value} for obj in cache.get("homepage_govuk_news")
        ]`

CamLamb:
Here are some templates left to move over:
- `about_us_new.html`
- `content_page_new.html`
- `content_with_related_topics_new.html`
- `news_home_new.html`
- `news_page_new.html`
- `section_home_alpha_order_new.html`
- `topic_new.html`
- `working_at_dit_home_new.html`

CamLamb (approved):

PR #756, opened 2024-09-26
INTR-305 - Move the events models/code into the content app - Step 1

david-okeke1337:
Incoming Changes:
- Moved events app models into a new events.py file in the content app.
- Split out the content/models.py file into base_content.py, blog.py and search.py

Todo: 
- Rewire existing usages.
- Delete old events app.
- Data migration*

Checklist:
- [ ] Tests are up to date

### DIFF
--- a/src/content/models/base_content.py
+++ b/src/content/models/base_content.py
@@ -1,0 +1,468 @@
+import html
+import logging
+from typing import Optional
+
+from django.contrib.auth import get_user_model
+from django.db import models
+from django.db.models import Q, Subquery
+from django.forms import widgets
+from django.utils import timezone
+from django.utils.html import strip_tags
+from content.models.search import SearchExclusionPageLookUp, SearchFieldsMixin, SearchPinPageLookUp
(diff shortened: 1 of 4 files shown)


src/content/model/__init__.py
CamLamb, line 1: Currently, Django is looking in the `models.py` file to find the model definitions. If we delete this file and rename the directory from `model` to `models` then Django will look in the `models/__init__.py` file for the definitions.

We can update the `__init__.py` file to import the models defined in each of the new files

src/content/model/events.py
```diff
@@ -0,0 +1,148 @@
+from datetime import datetime as dt
```
CamLamb, line 1: Once the changes from the previous message have been made, you can run `./manage.py makemigrations` which will create the event models in the `content` app.

When these models have been created, we can create a data migration to copy the data from the events app to the content app.

CamLamb:
Marked this PR as a draft as it is being actively worked on and is not ready for review/merging

PR #757, opened 2024-09-26
Update the country fact sheet page styling

CamLamb:


david-okeke1337 (approved):
Looks good, can't confirm the bookmark button at this moment due to a server error but styling changes look ok

PR #759, opened 2024-09-26
INTR-346 Home priority page order bug

CamLamb:
This PR adds a new panel type for selecting a page and preventing the ability to edit the page once it's set. A "Page Selector" instead of a "Page Chooser".

priority_pages

As you can see in the GIF above, the behaviour is as follows:
- Adding a new priority page allows for:
  - Selecting a page
  - Removing the page
  - Choosing a different page
- After saving the page as a draft you can still:
  - Selecting a page
  - Removing the page
  - Choosing a different page
- After publishing a page you can no longer change the page, you have to delete the priority page item and create a new instance

- [x] Tests are up to date

### DIFF
--- a/src/home/models.py
+++ b/src/home/models.py
@@ -14,7 +14,6 @@ from wagtail.admin.panels import (
     FieldPanel,
     InlinePanel,
     MultiFieldPanel,
-    PageChooserPanel,
 )
 from wagtail.models import PagePermissionTester
 from wagtail.snippets.models import register_snippet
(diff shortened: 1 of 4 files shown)


src/home/models.py
```diff
@@ -96,7 +96,7 @@ class HomePriorityPage(Orderable):
     )
 
     panels = [
-        PageChooserPanel("page", HOME_PRIORITY_PAGE_TYPES),
```
CamLamb, line 99: With the current `PageChooserPanel` in Wagtail, you can either always set it to read only or not. But oddly, the read only `PageChooserPanel` causes issue with the page saving as it doesn't display a hidden field.

alsyx (approved):

PR #766, opened 2024-10-03
INTR-342 Event page start/end dates

david-okeke1337:
- [ ] Tests are up to date

Created new event_start and event_end fields along with migration script.

Also updated the event page to dynamically show the correct event time depending on start/end day

/
image

/events/event/
image


src/events/utils.py
```diff
@@ -7,11 +7,17 @@
     from events.models import EventPage
 
 
-def get_event_date(event: "EventPage") -> str:
-    return event.event_date.strftime("%-d %B %Y")
+def get_event_start_date(event: "EventPage") -> str:
+    return f"{event.event_start.strftime("%-d %B %Y")}"
```
CamLamb: This doesn't actually need to be a f-string formatter
```suggestion
    return event.event_start.strftime("%-d %B %Y")
```

src/events/utils.py
```diff
@@ -7,11 +7,17 @@
...
 
 
-def get_event_time(event: "EventPage") -> str:
-    start_time = format_time(event.start_time)
-    end_time = format_time(event.end_time)
-    return f"{start_time} - {end_time}"
+def get_event_end_date(event: "EventPage") -> str:
+    return f"{event.event_end.strftime("%-d %B %Y")}"
```
CamLamb: Same here as there isn't anything being formatted in a string.
```suggestion
    return event.event_end.strftime("%-d %B %Y")
```

src/home/templatetags/cards.py
```diff
@@ -34,9 +39,21 @@ def page_to_card(page: NewsPage | EventPage, hide_shadow: bool = False):
...
+            event_post_title_time = (
+                f"{get_event_end_date(page)} {get_event_end_time(page)}"
+            )
+
         card_dict.update(
-            post_title_date=get_event_date(page),
-            post_title_time=get_event_time(page),
+            post_title_date=event_post_title_date,
+            post_title_time=event_post_title_time,
         )
```
CamLamb, line 57: Let's merge the `post_title_date` and `post_title_time` context variables to just be `post_title`
Then we can convert the logic above to create one long string.
The logic to generate the "Event datetime display string" can be moved into a function and replace the 4 existing methods:
- `get_event_end_date`
- `get_event_end_time`
- `get_event_start_date`
- `get_event_start_time`

This way we can use the new method to create the string for both the event page and in the `page_to_card` method.
The context and templates will likely need to be adjusted

CamLamb (changes requested):

src/events/utils.py
```diff
@@ -7,11 +7,17 @@
...
+    event_end_date = event.event_end.strftime("%-d %B %Y")
+    event_start_time = format_time(event.event_start.time())
+    event_end_time = format_time(event.event_end.time())
 
+    if event_start_date == event_end_date:
+        event_post_title = (
+            f"{event_start_date} <br>{event_start_time} - {event_end_time}"
+        )
+    else:
+        event_post_title = f"{event_start_date} {event_start_time} - <br>{event_end_date} {event_end_time}"
```
CamLamb, line 21: how does this look without the `<br>`?
Just wondering if this is needed so we can remove the `|safe` from the template.
If we want to keep the `<br>` then we can still remove the `|safe` and instead mark the output of this function as safe by using the [`mark_safe`](https://docs.djangoproject.com/en/5.1/ref/utils/#django.utils.safestring.mark_safe) function.

Note the risks of using `|safe` and `mark_safe` https://snyk.io/blog/preventing-xss-in-django/

src/events/templates/events/event_page.html
```diff
@@ -16,8 +16,7 @@
         <div class="dwds-card-wrapper">
             <div class="dwds-card dwds-information-panel">
                 <div>
-                    <h3>{{ event_date }}</h3>
-                    <h3>{{ event_time }}</h3>
+                    <h3>{{ event_title|safe }}</h3>
```
CamLamb, line 19: See other comment about `|safe`

CamLamb (changes requested):

CamLamb (approved):
Approved pending these final comments 😄

CamLamb (approved):

PR #773, opened 2024-10-09
add option to avoid sso on local

marcelkornblum:
Adds `dev_tools` app from IRAP project and configures it in the local developers settings to allow local devs to switch between users without going to SSO

### DIFF
--- a/src/config/urls.py
+++ b/src/config/urls.py
@@ -65,10 +65,11 @@ if settings.DEBUG:
     urlpatterns += staticfiles_urlpatterns()
     urlpatterns += static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
 
-    # Dev tools
-    urlpatterns += [
-        path("dev-tools/", include("dev_tools.urls", namespace="dev_tools"))
-    ]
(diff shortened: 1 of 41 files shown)


david-okeke1337 (approved):

src/config/urls.py
```diff
@@ -65,6 +65,11 @@
     urlpatterns += staticfiles_urlpatterns()
     urlpatterns += static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
 
+    # Dev tools
+    urlpatterns += [
+        path("dev-tools/", include("dev_tools.urls", namespace="dev_tools"))
+    ]
+
```
CamLamb, line 72: Might be worth putting this in a condition so that DEBUG doesn't assume `DEV_TOOLS_ENABLED`

```suggestion
    if settings.DEV_TOOLS_ENABLED:
        # Dev tools
        urlpatterns += [
            path("dev-tools/", include("dev_tools.urls", namespace="dev_tools"))
        ]

```

CamLamb (approved):

PR #779, opened 2024-10-16
INTR-384 Modifications to select and radio inputs design

nicopicchio:

image


src/events/templates/events/includes/listing_header.html
```diff
@@ -20,7 +20,7 @@ <h2>Filter events by</h2>
                                 {{ events_filters.form.location }}
                             </div>
                             <div class="content-stack">
-                                <label for="id_event_type">Event type</label>
+                                <label for="id_event_type">Event format</label>
```
alsyx, line 23: If we are changing this to be "Event format", it may be nice to update the options too. Instead of having "All types" it could be just "All" or "All formats"?

CamLamb: +1 for changing it to just `All`

CamLamb:
Screenshot 2024-10-17 at 13 52 13

I think the spacing to the left of the text in this box can be increased so that it matches the spacing on the top/bottom:

Screenshot 2024-10-17 at 13 53 00

src/events/templates/events/includes/listing_header.html
nicopicchio: Yeah I wanted to modify this too but I was not sure whether this needed to be discussed with content or not. I've made the modification and then we can reassess.

PR #788, opened 2024-10-25
[Poc] iCal subscribe progress

marcelkornblum:
iCal feed endpoint is exempted from SSO auth, though requires a user-specific token generated by visiting the `/ical` endpoint, which also gives access to a few URLs; the main ICS one and a couple of others designed to be "1 click subscribe" URLs for Outlook and Google. NB these are largely undocumented but seem the easiest way for users.

### DIFF
--- a/src/config/settings/developer.py
+++ b/src/config/settings/developer.py
@@ -1,4 +1,15 @@
+import environ
+import os
+
 from .base import *  # noqa
+
+
+# Read environment variables using `django-environ`, use `.env` if it exists
+env = environ.Env()
+env_file = os.path.join(PROJECT_ROOT_DIR, ".env")
+if os.path.exists(env_file):
+    env.read_env(env_file)
+env.read_env()
 
 DEBUG = True
 
@@ -38,7 +49,7 @@ except ModuleNotFoundError:
     ...
 
 
-DEV_TOOLS_ENABLED = True
+DEV_TOOLS_ENABLED = env.bool("DEV_TOOLS_ENABLED", True)
 
 if DEV_TOOLS_ENABLED:
     # remove Django Staff SSO Client for local login
--- a/src/events/views.py
+++ b/src/events/views.py
@@ -1,7 +1,9 @@
 from datetime import timedelta
 
 from django.http import HttpResponse
+from django.shortcuts import render
 from django.utils import timezone
+from django.utils.crypto import get_random_string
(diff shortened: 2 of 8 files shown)


src/config/settings/developer.py
```diff
@@ -1,5 +1,18 @@
...
 from .base import *  # noqa
 
+
+# Read environment variables using `django-environ`, use `.env` if it exists
+env = environ.Env()
+PROJECT_ROOT_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), "../../.."))
+env_file = os.path.join(PROJECT_ROOT_DIR, ".env")
+if os.path.exists(env_file):
+    env.read_env(env_file)
+env.read_env()
```
CamLamb, line 14: This feels a lil out of scope for this PR and probably wants some discussion.
While reading from a .env file is nice, it's not the current expected behaviour.

I like it as it will mean I don't have to restart a container to change the env var, but I will have to restart the container to make the runserver command restart.
Definitely worth discussing further.

src/events/views.py
```diff
@@ -55,7 +59,49 @@ def to_ical(event: EventPage) -> Event:
     return ical_event
 
 
+def get_user_token(user):
+    if existing_token := user.profile.ical_token:
+        return existing_token
+
+    token = get_random_string(length=80)
+    user.profile.ical_token = token
```
CamLamb, line 67: ❤️

src/events/views.py
```diff
@@ -55,7 +59,49 @@ def to_ical(event: EventPage) -> Event:
...
+# behind the VPN protection as well
 def ical_feed(request):
+    user = request.user
+    token = request.GET.get("tk", None)
+    if user.is_anonymous:
+        uuid = request.GET.get("u", None)
+        # some of our calendar link hacks double-parse the URL and can lose
+        # GEt vars after '&' so this is to get around that
+        if uuid is None and "." in token:
+            token, uuid = token.split(".")
```
CamLamb, line 99: This could deffo be split out into a function that explains the different reasons

CamLamb (approved):
Approving with comments, none of my comments block this from being merged

src/config/settings/developer.py
marcelkornblum: agreed

src/events/views.py
marcelkornblum: turns out encoding the links solves the problem and this hack isnt needed at all :)

PR #800, opened 2024-11-11
INTR-400: Build accordion dwds component

smanga24:
Accordion dwds component - Hide and show the subsections and all sections

### DIFF
--- a/src/dw_design_system/dwds/components/accordion.html
+++ b/src/dw_design_system/dwds/components/accordion.html
@@ -1,27 +1,32 @@
 {% load static %}
-<div class="dwds-accordions">
-    <div class="dwds-accordion-show-all-text"
-         style="border-bottom: 1px solid #b1b4b6;
-                line-height: 1.3157894737;
-                font-size: 1.1875rem;
-                color: #1d70b8;
-                font-weight: 400">
-        <span>Hide all sections</span>
+<div class="dwds-accordions dwds-accordion-template">
+    <div class="dwds-accordion-controls">
+        <button class="dwds-accordion-show-all">
+            <span class="dwds-accordion-show-all-chevron"></span>
+            <span class="dwds-accordion-show-all-text">Show all sections</span>
+        </button>
     </div>
 
     {% for item in items %}
-        <div class="dwds-accordion" style="color:#0b0c0c;">
-            <div class="dwds-accordion-header" style="font-family: arial, sans-serif;">
-                <h2>
-                    <span>{{ item.heading }}</span>
+        <div class="dwds-accordion-section">
+            <div class="dwds-accordion-section-header">
+                <h2 class="dwds-accordion-section-heading">
+                    <button class="dwds-accordion-section-button">
+
+                        <span class="dwds-accordion-section-heading-text">
+                            <span class="dwds-accordion-section-heading-text-focus">{{ item.heading }}</span>
+                        </span>
+                        <span class="dwds-accordion-section-toggle">
+                            <span class="dwds-accordion-section-toggle-focus">
+                                <span class="dwds-accordion-section-chevron"></span>
+                                <span class="dwds-accordion-section-toggle-text">Show</span>
+                            </span>
+                        </span>
+                    </button>
                 </h2>
             </div>
-            <div class="dwds-accordion-content"
-                 style="border-bottom: 1px solid #b1b4b6;
-                        -webkit-font-smoothing: antialiased;
-                        font-family: arial, sans-serif;
-                        font-weight:400;
-                        margin-top:0">
+
+            <div class="dwds-accordion-section-content">
                 <p>{{ item.content }}</p>
             </div>
 
--- a/src/dw_design_system/dwds/components/accordion.html
+++ b/src/dw_design_system/dwds/components/accordion.html
@@ -1,30 +1,30 @@
 {% load static %}
 <div class="dwds-accordions">
-    <div>
-        <button class="dwds-accordion-show-all-button">
-            <span class="dwds-accordion-show-all-chevron"></span>
-            <span class="dwds-accordion-show-all-text">Show all sections</span>
+    <div class="show-all-header">
+        <button>
+            <span class="show-all-chevron"></span>
+            <span class="show-all-text">Show all sections</span>
         </button>
     </div>
 
     {% for item in items %}
         <div class="dwds-accordion-section">
-            <div class="dwds-accordion-section-header">
+            <div class="section-header">
                 <h2>
                     <button>
-                        <span class="dwds-accordion-section-heading-text-focus">{{ item.heading }}</span>
-                        <span class="dwds-accordion-section-summary-text-focus">{{ item.summary }}</span>
-                        <span class="dwds-accordion-section-toggle">
-                            <span class="dwds-accordion-section-toggle-focus">
-                                <span class="dwds-accordion-section-chevron"></span>
-                                <span class="dwds-accordion-section-toggle-text">Show</span>
+                        <span class="text-large">{{ item.heading }}</span>
+                        <p class="text-small">{{ item.summary }}</p>
+                        <span class="section-toggle">
+                            <span class="section-toggle-focus">
+                                <span class="section-chevron"></span>
+                                <span class="section-text">Show</span>
                             </span>
                         </span>
                     </button>
                 </h2>
             </div>
 
-            <div class="dwds-accordion-section-content">
+            <div class="section-content">
                 <p>{{ item.content }}</p>
             </div>
 
--- a/src/dw_design_system/dwds/components/accordion.html
+++ b/src/dw_design_system/dwds/components/accordion.html
@@ -10,18 +10,14 @@
     {% for item in items %}
         <div class="dwds-accordion-section">
             <div class="section-header">
-                <h2>
-                    <button>
-                        <span class="text-large">{{ item.heading }}</span>
-                        <p class="text-small">{{ item.summary }}</p>
-                        <span class="section-toggle">
-                            <span class="section-toggle-focus">
-                                <span class="section-chevron"></span>
-                                <span class="section-text">Show</span>
-                            </span>
-                        </span>
-                    </button>
-                </h2>
+                <button>
+                    <span class="text-large">{{ item.heading }}</span>
+                    <p class="text-small">{{ item.summary }}</p>
+                    <span class="section-toggle">
+                        <span class="section-chevron"></span>
+                        <span class="section-text">Show</span>
+                    </span>
+                </button>
             </div>
 
             <div class="section-content">
--- a/src/dw_design_system/dwds/components/accordion.html
+++ b/src/dw_design_system/dwds/components/accordion.html
@@ -9,7 +9,7 @@
 
     {% for item in items %}
         <div class="dwds-accordion-section">
-            <div class="section-header" tabindex="0">
+            <div class="section-header">
                 <button>
                     <span class="text-large">{{ item.heading }}</span>
                     <p class="text-small">{{ item.summary }}</p>
--- a/src/dw_design_system/dwds/components/accordion.html
+++ b/src/dw_design_system/dwds/components/accordion.html
@@ -1,27 +1,32 @@
 {% load static %}
-<div>
+<div class="dwds-accordion">
     <div>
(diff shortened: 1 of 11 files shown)


src/dw_design_system/dwds/components/accordion.js
```diff
@@ -0,0 +1,43 @@
...
+            const accordionContent = accordion.getElementsByClassName("dwds-accordion-content")[0];
+            accordionHeader.addEventListener("click", function (e) {
+                toggleById(accordionContent);
+            });
+
+        });
+
+        // Add a click handler to the accordion header that toggles the visibility of all of the accordion contents
+        // Click handler calls `taggleAllSections` passing in each accordion element
+        const accordionsToggleAll = accordion.getElementsByClassName("dwds-accordion-show-all-text")[0];
```
CamLamb, line 40: `accordion` isn't available at this level, I think this should be `accordions`

smanga24: I suppose to comment out this code for now

CamLamb (commented):
Now that it's functional, there's a few more steps to consider:

- Run `make fix` which runs or formatters and linters
- Remove all inline styling into a component specific stylesheet
- Accessibility
  - Looking at the GOV UK component, there are some accessibility takeaways:
    - Hovering the interactive space makes the cursor change to show it is clickable
    - There is text that says show/hide
    - The "Show all sections" text is a button (likely for accessibility reasons)
    - There are aria tags: `aria-expanded` and `aria-label` in use:
      - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded
      - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label

src/dw_design_system/dwds/components/accordion.js
```diff
@@ -0,0 +1,39 @@
+function toggleById(elem) {
```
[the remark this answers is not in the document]
smanga24: Renamed to toggleSpecificSection()

src/dw_design_system/dwds/components/accordion.html
CamLamb, line 1: Can you add the option to add some optional text like this example from the GOV UK component?
Screenshot 2024-11-18 at 10 13 03

CamLamb: There's a lot of class names introduced here (I realise it's following from the GOV UK markup)

In our design system, we are trying to only add classes where it's needed, so if we can target a style by using the markup layout we aim to do that first.

Example:
If we have some HTML like so:
```html
<div class="class1">
    <p>Some text</p>
    <div>
        <img ... />
        <h2>A title</h2>
        <p>Some more text</P>
    </div>
</div>
```
Then instead of adding classes to everything, we can try to target most things just by the structure of the HTML:
```css
.class1 {
    > p {
        ...
    }
    
    > div {
        img {
            ...
        }

        h2 {
            ...
        }

        p {
            ...
        }
    }
}
```

There are scenarios where a class is needed, for example:

```HTML
<div class="class1">
    <div>...</div>
    <div>...</div>
    <div>...</div>
</div>
```

If you wanted to add some styling to a specific sub-div, it might not make sense to target the first/last/nth child, so adding a class in this scenario would be ok.

src/dw_design_system/dwds/components/accordion.scss
```diff
@@ -0,0 +1,154 @@
...
+.dwds-accordion-nav-chevron-down, .dwds-accordion-show-all-chevron-down{
+     transform: rotate(180deg);
+}
+
+.dwds-accordions {
+    border-bottom: 1px solid #b1b4b6;
+    margin-bottom: 20px;
+}
+
+@mixin accordion-chevron-hover-change {
```
CamLamb, line 109: We're trying to not add SCSS specific code where possible as their is a goal (potentially a long term one) to move away from SCSS.

It looks like the purpose of this mixin is to make the styling for the chevrons DRY which makes sense!
Although it looks like the styling for `.dwds-accordion-show-all-chevron` and `.dwds-accordion-section-chevron` are identical, so we could simplify that styling altogether (unless I've missed something)

src/dw_design_system/dwds/components/accordion.scss
```diff
@@ -0,0 +1,154 @@
...
+    color: #0b0c0c;
+    background: #0b0c0c;
+    &::after {
+        color: white;
+    }
+}
+
+.dwds-accordion-section-header:hover {
+    background-color: #eaeaea;
```
CamLamb, line 118: All of the colours in the stylesheet should be moved to the `variables.scss` file with named variables specific to their purpose

src/dw_design_system/dwds/components/accordion.scss
```diff
@@ -0,0 +1,154 @@
...
+     padding: 15px 15px 30px 15px;
+}
+
+.dwds-accordion-show-all-text{
+    margin-left: 5px;
+    vertical-align: middle;
+}
+
+.dwds-accordion-show-all {
+    font-family: GDS Transport, arial, sans-serif;
```
CamLamb, line 79: We can remove font things from our elements as the font used should be set globally and we only need to define font settings where we intentionally differ from the global font

src/dw_design_system/dwds/components/accordion.scss
```diff
@@ -0,0 +1,154 @@
...
+
+.dwds-accordion-section-heading {
+    padding: 0;
+    margin-top: 0;
+    margin-bottom: 0;
+}
+
+.dwds-accordion-section-heading-text {
+    display: block;
+    margin-bottom: 13px;
```
CamLamb, line 31: Where we define spacing (margin, padding, etc.) we should try to use the design system's spacing variables:
`--s*` from:
- `--s-5`
- `--s-4`
- `--s-3`
- `--s-2`
- `--s-1`
- `--s0`
- `--s2`
- `--s3`
- `--s4`
- `--s5`

CamLamb:
Most of the comment's I've added were in the original PR:
https://github.com/uktrade/digital-workspace-v2/pull/804

src/dw_design_system/dwds/components/accordion.scss
smanga24: I'm adding and removing "chevron down" in js code when chevron was clicked. So I need to identify that the click is on show-all-chevron or section-chevron.

src/dw_design_system/dwds/components/accordion.html
smanga24: 1. Added summary/sub section
2. Implemented nested structure and removed classes
3. Still using the classes wherever it's required.

src/dw_design_system/dwds/components/accordion.scss
smanga24: Removed mixin

smanga24:
> Most of the comment's I've added were in the original PR: #804

this is the main PR, we have not used other one any point in time.

src/dw_design_system/dwds/components/accordion.html
```diff
@@ -10,12 +10,10 @@
     {% for item in items %}
         <div class="dwds-accordion-section">
             <div class="dwds-accordion-section-header">
-                <h2 class="dwds-accordion-section-heading">
-                    <button class="dwds-accordion-section-button">
-
-                        <span class="dwds-accordion-section-heading-text">
-                            <span class="dwds-accordion-section-heading-text-focus">{{ item.heading }}</span>
-                        </span>
+                <h2>
```
CamLamb, line 13: I don't think we need a button here do we?

Also have you looked into the `<details>` component?
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details

It looks like that would handle the JS to toggle the hide/show for individual items, then we would only need to maintain the JS for the hide/show all logic.

Also the resulting HTML would be more semantic leading to a better chance of it being accessible to screen readers, etc.

src/dw_design_system/dwds/styles/variables.scss
```diff
@@ -66,9 +66,15 @@
   --color-icon: var(--dbt-blue);
   --color-hr: var(--gds-mid-grey);
   --color-selected-radio: var(--color-button-text-focus);
+  --color-background-accordion: var(--gds-light-grey);
+  --color-button-text-accordion: var(--gds-black);
+  --color-accordion-hover: #eaeaea;
 
   // Size
   --ratio: 1.25;
+  --s-8: calc(var(--s-7) / var(--ratio));
```
CamLamb, line 75: Do we need to go this small? was -5 too large?

src/dw_design_system/dwds/components/accordion.scss
```diff
@@ -1,42 +1,72 @@
...
-    border-top: 1px solid #b1b4b6;
-    border-bottom: 10px solid transparent;
-    color: #0b0c0c;
-    background: none;
-    text-align: left;
-    cursor: pointer;
-    -webkit-appearance: none;
+.dwds-accordions {
+    background-color: var(--color-background-accordion);
+    text-size-adjust: var(--width-max-percentage);
```
CamLamb, line 3: Do we need to use `text-size-adjust`?
We haven't used it anywhere else in the project and I don't think there is anything specific/unique about the accordion behaviour that would mean we need to reach for this

src/dw_design_system/dwds/components/accordion.scss
```diff
@@ -1,42 +1,72 @@
...
+        background: none;
+        text-align: left;
+        cursor: pointer;
+        -webkit-appearance: none;
+
+        font-size: var(--s2);
+        line-height: var(--ratio);
+
+        -webkit-font-smoothing: antialiased;
+        font-weight: 700;
```
CamLamb, line 31: The `--ratio` variable is used to define the ratio that the scaling variables increase/decrease by, so using it for the line height doesn't really make sense. There are line height variables available:
- `--font-lineheight-body`
- `--font-lineheight-headings`
- `--font-lineheight-interface`
Although I don't think we would need to worry too much about this if the HTML wasn't structured:
```HTML
<h2>
    <button>
        <span>Some heading text</span>
        <span>Some smaller text</span>
    </button>
</h2>
```
I think the above HTML layout would make more sense as:
```HTML
<button>
    <h2>Some heading text</h2>
    <p>Some smaller text</p>
</button>
```
That way the sizing of the text will come from the styles of the `h2` and `p` tag styling which is already set. And if it needs adjusting, then you can add those adjustments.

CamLamb: Note that the above comment is focussed just on the styling portion of this code, I still think that altering this component to be a `details` element might mean there is way less code for us to maintain.

src/dw_design_system/dwds/components/accordion.html
```diff
@@ -1,7 +1,7 @@
 {% load static %}
-<div class="dwds-accordions dwds-accordion-template">
-    <div class="dwds-accordion-controls">
-        <button class="dwds-accordion-show-all">
+<div class="dwds-accordions">
+    <div>
+        <button class="dwds-accordion-show-all-button">
```
CamLamb, line 4: The show all sections button doesn't seem to work when I tab onto it and try to activate it

smanga24: I'm sure you might be clicking outside of the scope It uses only text width

CamLamb: > when I tab onto it

This is when I am using my keyboard to test it.
I try to test components with both mouse and keyboard for a simple/brief accessibility check.

If I can press the `Tab` key to navigate to an element, I expect that element to perform the same way when I press enter as it does when I click on it.

src/dw_design_system/dwds/styles/variables.scss
smanga24: Removed and using -5. I'm doing like to like comparison.

src/dw_design_system/dwds/components/accordion.scss
smanga24: As we discussed, details element adds it's own look and feel. We stick to the button approche.

src/dw_design_system/dwds/components/accordion.html
CamLamb: After discussion this wasn't needed

CamLamb:
I've just pushed a load of commits that adjust the following:
- Move button styling into the `dwds-button` component under a new `--clear` stlye
- Update the component to use the "stack" concept
- Swap out to the correct CSS variables
- Reorder the accordion styles

CamLamb (approved):

PR #802, opened 2024-11-13
Add a crude method of doing an exact search

CamLamb:


nicopicchio (approved):

src/search/search.py
```diff
@@ -43,6 +44,10 @@ def get_queryset(self):
         return self.model.objects.all()
 
     def build_query(self, query_str, *args, **kwargs):
+        if query_str.startswith('"') and query_str.endswith('"'):
+            if new_query_str := query_str[1:-1]:
+                # If wrapped in quotes, we want to search for an exact match.
+                return Phrase(new_query_str)
         return CustomQueryBuilder.get_search_query(self.model, query_str)
```
CamLamb, line 51:
```suggestion
        if query_str.startswith('EXACT_SEARCH '):
            if new_query_str := query_str[13:]:
                # If wrapped in quotes, we want to search for an exact match.
                return Phrase(new_query_str)
        return CustomQueryBuilder.get_search_query(self.model, query_str)
```

marcelkornblum (approved):

PR #805, opened 2024-11-18
INTR-399 Update button styling

david-okeke1337:
- [x] Tests are up to date

This PR updated the existing dwds button styles to have a thicker border at the bottom as per the designs 

/dwds/styles
image

### DIFF
--- a/src/dw_design_system/dwds/styles/variables.scss
+++ b/src/dw_design_system/dwds/styles/variables.scss
@@ -93,6 +93,10 @@
 
   // Borders and element widths
   --border-width: 3px;
+  --border-width-top-navigation: 2px;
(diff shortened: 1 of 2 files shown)


src/dw_design_system/dwds/styles/variables.scss
david-okeke1337, line 1: Note: I left border-width alone as it's used elsewhere (layout-primitives.scss) and didn't want to mess with the styling on .box there. Let me know if you think that's ok. Also, I'm not sure what naming convention would be preffered, should I change the variable name to something like dwds-button-border-width?

CamLamb (approved):
Looks good to me!

PR #822, opened 2024-11-28
Bug: Events in december don't show

CamLamb:


src/events/models.py
```diff
@@ -78,6 +80,11 @@ def get_context(self, request, *args, **kwargs):
             page_title_prefix = "What happened in"
 
         # Filtering events
+        print("CAM WAS HERE")
+        print(events)
+        print(month_start)
+        print(month_end)
+        print(request.GET)
```
nicopicchio, line 87: do you want to leave these `print()` in?

nicopicchio (approved):

PR #831, opened 2024-12-04
Reduce home page DB query count

CamLamb:
Reduces the number of queries on the home page by getting the images from the DB earlier


src/home/models.py
```diff
@@ -299,7 +304,8 @@ def get_context(self, request, *args, **kwargs):
         # Load the priority pages, preserving the order.
         priority_pages = [
             p.specific
-            for p in ContentPage.objects.filter(id__in=priority_page_ids)
+            for p in ContentPage.objects.select_related("preview_image")
```
smanga24, line 307: If the relationship is OneToMany, then prefetch_related would be a good choice

CamLamb: It is in reference to this field:
https://github.com/uktrade/digital-workspace-v2/blob/0b96d5e0738cbc6c0114dcc2408a761f7b6e4316/src/content/models.py#L354

Which means it is a many-to-one, as in many pages to one image.

smanga24 (approved):

CamLamb:
Currently blocked by the [Django toolbar PR](https://github.com/uktrade/digital-workspace-v2/pull/825)

PR #839, opened 2024-12-10
INTR-466: Implemented feedback component

smanga24:
- [ ] Tests are up to date


src/dw_design_system/dwds/components/feedback.js
```diff
@@ -0,0 +1,9 @@
+document.addEventListener('DOMContentLoaded', function () {
+    const fbButton = document.querySelector('.feedback-btn');
+    const fbElement = document.querySelector('.feedback-section details');
+    fbButton.addEventListener('click', function(event) {
+        event.preventDefault();
+        fbElement.open = true;
+        fbElement.scrollIntoView({ behavior: 'smooth' });
```
CamLamb, line 7: Instead of just scrolling the form into view, we should focus the first form element as this will inform screen readers that something has changed on the page

src/core/templates/base.html
```diff
@@ -100,7 +100,9 @@
                 {% block main_content %}
                 {% endblock main_content %}
                 {% block page_problem %}
-                    {% include "includes/problem_found.html" %}
+                    <section class="feedback-section">
+                        {% include "includes/problem_found.html" %}
+                    </section>
```
CamLamb, line 105: Why are we adding the feedback section here?

src/content/templates/content/content_page.html
```diff
@@ -20,3 +20,9 @@
     {% endif %}
     {{ block.super }}
 {% endblock post_primary_content %}
+
+{% block secondary_content %}
+    <section class="feedback-section">
```
CamLamb, line 25: Should the wrapping HTML be defined in the template since it is being used everywhere the template is included?

src/dw_design_system/dwds/components/feedback.html
```diff
@@ -0,0 +1,9 @@
+<div class="feedback-section">
+    <button class="feedback-btn">
+        <div class="feedback-icon">{% include "dwds/icons/feedback.html" %}</div>
```
CamLamb, line 3: The `feedback-icon` class isn't used anywhere so isn't needed

src/dw_design_system/dwds/components/feedback.html
```diff
@@ -0,0 +1,9 @@
+<div class="feedback-section">
```
CamLamb, line 1: This component could be named `dwds-sidebar-feedback` since other components are prefixed `dwds`

src/dw_design_system/dwds/components/feedback.html
```diff
@@ -0,0 +1,9 @@
...
+    <button class="feedback-btn">
```
CamLamb, line 2: The class name `feedback-btn` isn't needed as it can be targeted like so:
`.dwds-sidebar-feedback > button`

src/home/templates/home/home_page.html
```diff
@@ -93,6 +93,9 @@
                     {% include "dwds/components/link_list.html" with title="Quick links" list=quick_links %}
                 </div>
             {% endif %}
+            <div class="feedback-card">
+                {% include 'dwds/components/feedback.html' with title="Give feedback" description="Did you find what you were looking for?" %}
+            </div>
```
CamLamb, line 98: The ticket mentions creating a template tag for rendering the sidebar so that we don't have to configure each template over and over.

src/core/templates/base.html
smanga24: I'm using the section's class to target the feed back form in java script code

src/dw_design_system/dwds/components/feedback.js
smanga24: I think scroll into view shows the whole feedback form on the screen and thought should be good option. But i will change the code to focus on first element form

src/dw_design_system/dwds/components/feedback.html
smanga24: Removed

src/dw_design_system/dwds/components/feedback.html
smanga24: Renamed as suggested

src/dw_design_system/dwds/components/feedback.html
smanga24: Removed the button's class

src/content/templates/content/content_page.html
smanga24: created template tag

src/home/templates/home/home_page.html
smanga24: created template tag

src/dw_design_system/dwds/components/feedback.js
smanga24: changed to focus

src/core/templatetags/feedback.py
```diff
@@ -0,0 +1,14 @@
+from django import template
+
+
+register = template.Library()
+
+
+@register.inclusion_tag("dwds/components/feedback.html")
+def feedback_sidebar(
```
CamLamb, line 8: This is a good start for the template tag, but as described in the ticket we want to make something that can be built upon and extended:

> The sidebar will be made up of many sections. Each section will contain components that will decide if they need to be in use for the current page, if so they will be rendered in a given order.

So I was thinking that this should be a generic `sidebar` template tag that defines some kind of configuration that can be easily and quickly built upon.

A rough code idea that probably doesn't work but I am using to get the idea across:
```python
SIDEBAR_SECTIONS = [
    {
        "id": "primary_page_actions"
        "parts": [],
    },
    {
        "id": "secondary_page_actions"
        "parts": [
            {
                "id": "feedback",
                "is_visible": None, # A callable function that returns a boolean to show if the feedback should be visible or not
                "render": None, # A callable function that returns the rendered template
            }
        ],
    },
]
```
After writing out my thoughts above, it might make sense for the `parts` to be a list of custom Python classes that have the functions defined on them instead of dictionary objects.

Then the render logic for the sidebar could be:
- Loop over each sidebar section
- Loop over each part in the sidebar section
- Check to see if the part should be visible given the current URL/page/logged in user/etc.
- If the above is True, render the part

smanga24: Implemented to extend this basic functionality to add more sidebar options

CamLamb (approved):

PR #846, opened 2024-12-31
INTR-471: Return None if no alert found

smanga24:
- [ ] Tests are up to date

### DIFF
--- a/src/core/templatetags/sidebar.py
+++ b/src/core/templatetags/sidebar.py
@@ -41,20 +41,21 @@ class SidebarSection:
 class SiteAlert(SidebarPart):
     template_name = "tags/sidebar/site_alert.html"
 
+    def __init__(self, context):
+        super().__init__(context)
+        self.current_alert = SiteAlertBanner.objects.filter(activated=True).first()
+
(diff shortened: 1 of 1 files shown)


src/core/templatetags/sidebar.py
```diff
@@ -49,10 +49,12 @@ def is_visible(self, *args, **kwargs):
...
-        return {
-            "banner_text": current_alert.banner_text,
-            "banner_link": current_alert.banner_link,
-        }
+        if current_alert:
+            return {
+                "banner_text": current_alert.banner_text,
+                "banner_link": current_alert.banner_link,
+            }
+        return None
```
CamLamb, line 57: I think the right approach for this might be to move the database call:
`current_alert = SiteAlertBanner.objects.filter(activated=True).first()`

Into the `__init__` method under `self.current_alert` that way we can determine if the part should be visible or not based on if there is an alert to show.

smanga24: Done the changes, Alert is visible only when exists

CamLamb (approved):

PR #850, opened 2025-01-02
INTR:471 Move bookmark to primary page actions

smanga24:
- [ ] Tests are up to date


src/core/templatetags/sidebar.py
```diff
@@ -98,6 +98,23 @@ def get_part_context(self):
...
+        page = self.context.get("self")
+        if isinstance(page, HomePage):
+            return False
+        return True
+
+    def get_part_context(self):
+        return {
+            "user": self.context.get("user"),
+            "page": self.context.get("self"),
+            "csrf_token": self.context["csrf_token"],
```
CamLamb, line 114: I've looked at the flow of this and I think we can likely remove the `bookmark_page_input` template tag and move the contents of the `interactions/bookmark_page_input.html` template into your new wrapper template. Then we can move the logic from the `bookmark_page_input` template tag into this `get_part_context` function

CamLamb: Also, looking at all of the HTML in the templates, we will likely want to ensure the following is all in the context:
- `is_bookmarked`
- `request`
- `post_url`
- `page`
- `csrf_token`

smanga24: Done the changes

CamLamb:
The backend changes are looking great!
Please can the frontend be updated to more closely match the designs:
image

src/core/templatetags/sidebar.py
```diff
@@ -98,6 +100,29 @@ def get_part_context(self):
         }
 
 
+class Bookmarks(SidebarPart):
```
CamLamb, line 103: Can we rename this to `Bookmark` since it is the functionality to bookmark this page, therefore relates to one bookmark :)

CamLamb (approved):

PR #865, opened 2025-01-14
Only show bookmark icon on pages

CamLamb:
The bookmarks sidebar part is returning that it should be visible on pages that are NOT a Wagtail Page.

Before:
Screenshot 2025-01-14 at 16 57 43

After:
Screenshot 2025-01-14 at 16 58 57


nicopicchio (approved):

smanga24 (commented):
last 3 lines can be simplified with return isinstance(page, Page)

PR #869, opened 2025-01-15
INTR-506: create, update and delete reactions for a given page and user

smanga24:
- [ ] Tests are up to date

### DIFF
--- a/src/interactions/services/reactions.py
+++ b/src/interactions/services/reactions.py
@@ -1,0 +1,18 @@
+from content.models import BasePage
+from interactions.models import Reaction, ReactionType
+from news.models import NewsPage
+from user.models import User
+
+
+def manage_reaction(user: User, page: BasePage, reaction_type: str):
+
+    if not isinstance(page, NewsPage):
+        raise ValueError("The page must be a NewsPage.")
+
+    if not reaction_type or reaction_type not in ReactionType.values:
+        Reaction.objects.filter(user=user, page=page).delete()
+        return None
+
+    return Reaction.objects.update_or_create(
+        user=user, page=page, defaults={"type": reaction_type}
+    )
--- a/src/interactions/services/reactions.py
+++ b/src/interactions/services/reactions.py
@@ -1,10 +1,11 @@
-from content.models import BasePage
+from wagtail.models import Page
+
 from interactions.models import Reaction, ReactionType
 from news.models import NewsPage
 from user.models import User
 
 
-def manage_reaction(user: User, page: BasePage, reaction_type: str):
+def manage_reaction(user: User, page: Page, reaction_type: str):
 
     if not isinstance(page, NewsPage):
         raise ValueError("The page must be a NewsPage.")
--- a/src/interactions/services/reactions.py
+++ b/src/interactions/services/reactions.py
@@ -5,15 +5,20 @@ from news.models import NewsPage
 from user.models import User
 
 
-def manage_reaction(user: User, page: Page, reaction_type: str):
+def react_to_page(user: User, page: Page, reaction_type: str | None):
 
     if not isinstance(page, NewsPage):
         raise ValueError("The page must be a NewsPage.")
 
-    if not reaction_type or reaction_type not in ReactionType.values:
+    if reaction_type is None:
         Reaction.objects.filter(user=user, page=page).delete()
         return None
 
-    return Reaction.objects.update_or_create(
+    if reaction_type not in ReactionType.values:
+        raise ValueError(f"{reaction_type} is not a valid reaction type.")
+
+    reaction, created = Reaction.objects.update_or_create(
         user=user, page=page, defaults={"type": reaction_type}
     )
+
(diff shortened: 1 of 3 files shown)


src/interactions/services/reactions.py
```diff
@@ -0,0 +1,19 @@
+from wagtail.models import Page
+
+from interactions.models import Reaction, ReactionType
+from news.models import NewsPage
+from user.models import User
+
+
+def manage_reaction(user: User, page: Page, reaction_type: str):
```
CamLamb, line 8: Functions in the service layer should be named to be more task oriented with focussed and clear names. The word "Manage" could mean a few things. Perhaps something like `react_to_page` or something similar

src/interactions/services/reactions.py
```diff
@@ -0,0 +1,19 @@
...
+
+    if not isinstance(page, NewsPage):
+        raise ValueError("The page must be a NewsPage.")
+
+    if not reaction_type or reaction_type not in ReactionType.values:
+        Reaction.objects.filter(user=user, page=page).delete()
+        return None
```
CamLamb, line 15: I think there are 2 different scenarios that we want to manage here:
- If the `reaction_type` is None we want to clear the reaction from the database
- If the `reaction_type` is not an allowed Reaction type value raise an issue

```suggestion
    if reaction_type is None:
        Reaction.objects.filter(user=user, page=page).delete()
        return None

    if reaction_type not in ReactionType.values:
        raise ValueError(f"{reaction_type} is not a valid ReactionType")
```

src/interactions/services/reactions.py
CamLamb, line 8: The type for `reaction_type` should be `str | None` based on the code within the method.

src/interactions/test/services/test_reactions.py
```diff
@@ -0,0 +1,74 @@
...
+def about_page():
+    return AboutUs.objects.create(
+        title="About Us",
+        depth=4,
+        path="page1",
+    )
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize("reaction_type", ALL_REACTION_TYPES)
```
CamLamb, line 38: I really like this use of `parametrize`!

src/interactions/test/services/test_reactions.py
```diff
@@ -0,0 +1,74 @@
...
+def test_manage_reaction_create(user, news_page, reaction_type):
+    manage_reaction(user, news_page, reaction_type)
+    assert_reaction_type(user, news_page, reaction_type)
```
CamLamb, line 41: Since `manage_reaction` returns the `Reaction` object or `None` we should use that object instead of doing another DB call in the `assert_reaction_type` method

CamLamb, line 41: Instead of having the `assert_reaction_type` function that doesn't really hold any complexity, it can just be replaced with the following:

```python
reaction = manage_reaction(user, news_page, reaction_type)
assert reaction.type == reaction_type
```

src/interactions/test/services/test_reactions.py
```diff
@@ -0,0 +1,74 @@
...
+def assert_reaction_type(user, page, expected_type):
+    reaction = Reaction.objects.get(user=user, page=page)
+    assert reaction.type == expected_type
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize("reaction_type", ALL_REACTION_TYPES)
+# Test for user deselects reaction
+def test_manage_reaction_delete(user, news_page, reaction_type):
+    create_reaction(user, news_page, reaction_type)
```
CamLamb, line 61: Instead of a create reaction function, we can make it a fixture 😄

src/interactions/test/services/test_reactions.py
```diff
@@ -0,0 +1,74 @@
...
+from interactions.models import Reaction, ReactionType
+from interactions.services.reactions import (
+    manage_reaction,
+)
+from news.models import NewsPage
+
+ALL_REACTION_TYPES = ReactionType.values
+
+
+@pytest.fixture
```
CamLamb, line 14: We should move fixtures into a `conftest.py` file in the `src/interactions/test` directory
There are other examples of `conftest.py` files in the project

src/interactions/services/reactions.py
smanga24: updated to react_to_page

smanga24: Added None

src/interactions/test/services/test_reactions.py
smanga24: Done

smanga24: removed assert_reaction_type fun

src/interactions/test/services/test_reactions.py
smanga24: Make it fixture

CamLamb (approved):
LGTM!

PR #870, opened 2025-01-15
INTR-473 - Sidebar primary page actions comment

david-okeke1337:
- [ ] Tests are up to date

Testing:

As a superuser, ensure the "new_sidebar" flag is created (/django-admin/core/featureflag/) to see the sidebar.

Create/Navigate to a news page
image

Clicking on the comment icon should anchor the page to the comment box.


src/interactions/templates/interactions/comment_page_input.html
CamLamb, line 1: Comments aren't a concept held by the `interactions` app. Therefore I think that all of this can be moved into the `src/core/templates/tags/sidebar/parts/comment.html` template. And this template can be removed

src/core/templatetags/sidebar.py
```diff
@@ -170,6 +171,27 @@ def get_part_context(self):
...
 
+class Comment(SidebarPart):
+    template_name = "tags/sidebar/parts/comment.html"
+
+    def is_visible(self):
+        request = self.context["request"]
+        page = self.context.get("self")
+        if not flag_is_active(request, "new_sidebar"):
+            return False
+        return bool(isinstance(page, NewsPage) and page.allow_comments)
```
CamLamb, line 182: Just a slight suggestion on reordering the function so it reads a little easier. The principle here is only defining the variables as we need them.
```suggestion
    def is_visible(self):
        request = self.context["request"]
        if not flag_is_active(request, "new_sidebar"):
            return False
        
        page = self.context.get("self")
        return bool(isinstance(page, NewsPage) and page.allow_comments)
```

CamLamb (commented):
Just a few comments, other than that it's looking great!
Let me know when the comments are addressed and I will re-review/approve!

CamLamb (approved):

PR #871, opened 2025-01-16
Use management command to run debugpy

smanga24:
- [ ] Tests are up to date

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Project documentation is available [here](https://uktrade.github.io/digital-work
 
 Add environment variable in your .env file
 
-    ENABLE_DEBUGPY=True
+    DEBUGPY_ENABLED=True
 
 Create launch.json file inside .vscode directory
 
--- a/README.md
+++ b/README.md
@@ -19,8 +19,10 @@ Create launch.json file inside .vscode directory
                 "name": "Python: Remote Attach (DebugPy)",
                 "type": "debugpy",
                 "request": "attach",
-                "port": 5678,
-                "host": "localhost",
+                "connect": {
+                    "host": "localhost",
+                    "port": 5678
+                },
                 "pathMappings": [
                     {
                         "localRoot": "${workspaceFolder}",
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -17,7 +17,7 @@ Create launch.json file inside .vscode directory
         "configurations": [
             {
                 "name": "Python: Remote Attach (DebugPy)",
-                "type": "python",
+                "type": "debugpy",
                 "request": "attach",
                 "port": 5678,
```
CamLamb, line 22: I think the port and host have to go like this when using `debugpy` as the type:
```
"connect": {
    "host": "0.0.0.0",
    "port": 5678,
},
```

pyproject.toml
```diff
@@ -123,7 +123,7 @@ omit = ["*/migrations/*", "*/test/*", "e2e_tests/*"]
 [tool.ruff]
 src = ["src"]
 lint.extend-select = ["B", "S", "I"]
-lint.ignore = ["B904", "E501", "N818", "S101"]
+lint.ignore = ["B904", "E501", "N818", "S101", "S104"]
```
CamLamb, line 126: We shouldn't globally ignore this rule:
https://docs.astral.sh/ruff/rules/hardcoded-bind-all-interfaces/

Instead we can ignore just the line it is raising an error on (if we are happy with the line) using `# noqa: S104`

CamLamb: According to the rule, we should probably try `127.0.0.1` instead of `0.0.0.0`. Try that and see if the server still runs and can be connected to?

smanga24: I have tried using 127.0.0.1 but it's not working

smanga24: Updated with inline ignore

README.md
smanga24: Tried with 0.0.0.0 but not working

CamLamb: This comment was highlighting that the values `port` and `host` should be inside a `connect` object

CamLamb (approved):

PR #874, opened 2025-01-20
INTR-511 New Networks homepage

nicopicchio:
- [ ] Tests are up to date


src/interactions/templates/interactions/bookmark_page_input.html
CamLamb, line 1: Has this change been tested with the `new_sidebar` flag enabled?

src/home/templatetags/cards.py
```diff
@@ -54,7 +55,7 @@ def render(self, *args, **kwargs) -> SafeString:
 
 @register.simple_tag
 def page_to_engagement(
-    page: NewsPage | EventPage, hide_updated: bool = False
+    page: NewsPage | EventPage | NetworksHome, hide_updated: bool = False
```
CamLamb, line 58: When is this function being called with a `NetworksHome`?

It looks like you call it with a `Network`

src/networks/templates/networks/includes/network_listing_section.html
CamLamb, line 1: This template isn't being reused anywhere so we can move it's contents into the `networks_home.html`

src/home/templatetags/cards.py
nicopicchio: yes, I forgot to change that!

PR #894, opened 2025-01-27
INTR-526: Add reaction icons to the design system

smanga24:
- [ ] Tests are up to date


src/dw_design_system/dwds/icons/celebrate.html
```diff
@@ -0,0 +1,18 @@
+<svg class="celebrate-icon hover-light"
```
CamLamb, line 1: Can all icons have the `content-icon` class please 😄

CamLamb, line 1: The `hover-light` class should be added based on some variables so that we can toggle the logic based on usage
The same is true for `dark` and `hover-dark`.

I would just copy this from the bookmark icon:
```
{% if dark %}dark{% endif %} {% if hover_light %}hover-light{% endif %} {% if hover_dark %}hover-dark{% endif %}
```

This behaviour should also be consistent against all icons,

src/dw_design_system/dwds/icons/dislike.html
```diff
@@ -0,0 +1,10 @@
+<svg class="dislike-icon hover-light"
+     width="43"
+     height="43"
+     viewBox="0 0 43 43"
```
CamLamb, line 4: Is there a reason that this icon has a height and width of `43` and not `42` like the celebrate icon (and others already in the codebase)

Please don't change these to 42 unless you know for sure that this is the right thing to do as this might result in misalignments and clipping of the SVG.

I'm wanting to know if there is a reason as we might need to discuss with Cheryl the icon consistency across the project to make sure we don't accidentally build issues with sizing/alignment, etc.

This applies to all of the icons that differ

src/dw_design_system/dwds/icons/celebrate.scss
CamLamb, line 1: It would be good if these `scss` files could be as close to the same formatting and ordering as the existing ones with similarly named variables so that we have a more consistent codebase:
- `src/dw_design_system/dwds/icons/comment.scss`
- `src/dw_design_system/dwds/icons/bookmark.scss`

Note the variables names start with the component name`--bookmark-icon` and include a name that describes it's purpose: `background-border-color`

src/dw_design_system/dwds/icons/unhappy.scss
```diff
@@ -0,0 +1,62 @@
...
+    --foreground-circle-stroke:var(--color-foreground-black);
+    --curved-mouth-stroke:var(--color-foreground-black);
+    --foreground-color:var(--color-foreground-black);
+
+     &.dark {
+          --background-color:var(--color-background-very-light-grey);
+          --background-shade:var(--color-dark-purple);
+          --foreground-circle-color:var(--color-dark-purple);
+          --foreground-circle-stroke:'';
+          --curved-mouth-stroke:'';
```
CamLamb, line 14: I think you can use `unset` here. If you can, that would be preferred IMO as it is explaining more about what is happening

src/dw_design_system/dwds/icons/celebrate.scss
smanga24: Sure I will update the variable names, i thought these names are very longer. But I will follow the same approach as bookmark

src/dw_design_system/dwds/icons/dislike.html
smanga24: Done

src/dw_design_system/dwds/icons/celebrate.html
smanga24: Done

smanga24: Done

src/dw_design_system/dwds/icons/celebrate.scss
smanga24: Done

src/dw_design_system/dwds/icons/unhappy.scss
smanga24: Done

src/dw_design_system/dwds/icons/dislike.html
CamLamb: Looking at the designs (see slack message)
In the purple box on the far right are all of the icons and they're 42x42

So the change is right, but can you confirm that the SVG code used is from these icons in this location of the design file?

smanga24: I have taken all svg images from the specified location. Now like and dislike icons are not 42X42

CamLamb: Cheryl has updated the icons in the design (see slack)

CamLamb (approved):

PR #900, opened 2025-01-31
Fix the failing builds

CamLamb:
The build step is failing as it needs to run `collectstatic`. It does this outside of docker, so there is no env set.
This means that it needs to load a `.env` file. It doesn't look like there is a way to configure paketo buildpack to load this in, so I have opted to move this logic into the `config.settings.build` to keep the nasty code near to the problem.


marcelkornblum (approved):
Minor stylistic quibble aside, I think this works. It does make me wonder if there isn't a more paketo-friendly way of injecting everything from the file into the env; maybe in a shell script that runs as the environment prepares itself?

PR #902, opened 2025-02-03
WIP - INTR-539 GTM reactions

david-okeke1337:
- [ ] Tests are up to date


src/core/templates/includes/gtm_datalayer_info.html
```diff
@@ -8,6 +8,7 @@
     "page_topics": "{% for topic in self.topics.all %}{{ topic.topic.title }} {% endfor %}",
     "page_tags": "{% for tag in self.tag_set %}{{ tag.name }} {% endfor %}",
     "page_content_owner": "{{ self.content_owner.slug|default:"NA"}}",
+    "page_reactions": {% if not self.reactions is not None %}{{ self.reactions }}{% else %}"NA"{% endif %},
```
nicopicchio, line 11: I think this needs to be wrapped in quotation marks.
```
"{% if not self.reactions is not None %}{{ self.reactions }}{% else %}"NA"{% endif %}"
```
Also, not sure if it could be simplified along the same lines of line 10?
```
"page_reactions": "{{ self.reactions|default:'NA' }}",
```

david-okeke1337:
Todo: fix reaction block not getting set after on subsequent button presses
image

PR #904, opened 2025-02-04
INTR-528: Add reactions to the news page

smanga24:
- [ ] Tests are up to date


src/interactions/views.py
```diff
@@ -52,3 +54,54 @@ def bookmark_index(request, *args, **kwargs):
             "bookmarks": bookmarks_service.get_bookmarks(request.user),
         },
     )
+
+
+@require_http_methods(["GET", "POST"])
+def react_to_page(request, *args, pk, **kwargs):
```
CamLamb, line 60: I managed to get this view working nicely for us yesterday, but there's definitely some tidy up we can do.
Maybe move some of the context generation before `if request.method == "POST":` if it is shared between both return types, etc.

smanga24: Simplified context to use both in post and get methods

CamLamb (approved):

PR #905, opened 2025-02-04
Update the vertical menu to load current page headings

CamLamb:
Passing a truthy value for `enable_current_page_menu ` to the vertical menu component should result in a sub menu below the current page link which is made up of the headings from the page content


src/dw_design_system/dwds/components/menu_vertical.html
```diff
@@ -11,3 +20,30 @@
...
+                    return;
+                }
+
+                let heading_link = document.createElement('a');
+                heading_link.href = '#' + heading_id;
+                heading_link.textContent = heading_text;
+
+                let heading_item = document.createElement("li");
+                heading_item.appendChild(heading_link);
+                table_of_contents.appendChild(heading_item);
```
nicopicchio, line 45: Should we not use `const` rather then `let` as there we are not reassigning the variables?

CamLamb: Good catch!

nicopicchio (approved):
LGTM 🚀

PR #919, opened 2025-02-10
INTR-558: Rename network type option - development to networks

smanga24:
- [ ] Tests are up to date


CamLamb (commented):
This will just update the choices, a data migration is required to update the existing instances of `Network` so that their value changes from the old one to the new one.

smanga24:
> This will just update the choices, a data migration is required to update the existing instances of `Network` so that their value changes from the old one to the new one.

Done

src/networks/migrations/0008_alter_network_network_type.py
```diff
@@ -0,0 +1,43 @@
+# Generated by Django 5.1.5 on 2025-02-10 14:11
+
+from django.db import migrations, models
+
+
+def update_network_type(apps, schema_editor):
```
CamLamb, line 6: I've just tested this locally and it doesn't look to have worked

src/networks/migrations/0008_alter_network_network_type.py
```diff
@@ -0,0 +1,43 @@
...
+                        "Professional Networks and Skills",
+                    ),
+                    ("social_and_community", "Social and Community"),
+                    ("volunteering", "Volunteering"),
+                ],
+                max_length=50,
+                null=True,
+            ),
+        ),
+        migrations.RunPython(update_network_type),
```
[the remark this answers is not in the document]
smanga24: Rollback changes added

src/networks/migrations/0008_alter_network_network_type.py
smanga24: Let me check it again locally

src/networks/migrations/0008_alter_network_network_type.py
```diff
@@ -0,0 +1,79 @@
+# Generated by Django 5.1.5 on 2025-02-10 14:11
+from django.db import migrations, models
+from django.contrib.contenttypes.models import ContentType
+from wagtail.models import Revision
```
CamLamb, line 4: In migrations, models should not be imported like this, they are added inline like so:
`Network = apps.get_model("networks", "Network")`

This is because the model in the code will be different to the model that you are accessing mid way through making changes to the database.

Heres some helpful links for further reading on the topic and better explanations:
https://docs.djangoproject.com/en/5.1/topics/migrations/#historical-models
https://codereview.doctor/features/django/best-practice/avoid-model-imports-in-migrations

src/networks/migrations/0008_alter_network_network_type.py
```diff
@@ -0,0 +1,87 @@
...
+    ContentType = apps.get_model("contenttypes", "ContentType")
+    content_type = ContentType.objects.get(app_label="networks", model="network")
+
+    update_network_page_and_revision(
+        Network.objects.all(),
+        old_network_type,
+        new_network_type,
+        Revision,
+        content_type,
+    )
```
CamLamb, line 38: Since the `update_network_page_and_revision` function is only called once, there's no need to separate the logic, we can keep it all in one method.

```suggestion
def run_update_network_type(apps, old_network_type, new_network_type):
    Network = apps.get_model("networks", "Network")
    Revision = apps.get_model("wagtailcore", "Revision")

    ContentType = apps.get_model("contenttypes", "ContentType")
    content_type = ContentType.objects.get(app_label="networks", model="network")

    for network_page in Network.objects.all():
        if network_page.network_type == old_network_type:
            network_page.network_type = new_network_type
            network_page.save(update_fields=["network_type"])

        revisions = Revision.objects.filter(
            content_type=content_type, object_id=network_page.id
        )
        for revision in revisions:
            if revision.content["network_type"] == old_network_type:
                revision.content["network_type"] = new_network_type
                revision.save(update_fields=["content"])
```

This also ends up being less lines of code which is nice 😄

CamLamb (commented):
Other than the one comment, looking good! I will test locally!

src/networks/migrations/0008_alter_network_network_type.py
smanga24: Done

src/networks/migrations/0008_alter_network_network_type.py
```diff
@@ -0,0 +1,75 @@
...
+OLD_NETWORK_TYPE = "professional_development_and_skills"
+NEW_NETWORK_TYPE = "professional_networks_and_skills"
+
+
+def run_network_type_update(apps, old_network_type, new_network_type):
+    Network = apps.get_model("networks", "Network")
+    Revision = apps.get_model("wagtailcore", "Revision")
+
+    ContentType = apps.get_model("contenttypes", "ContentType")
+    content_type = ContentType.objects.get(app_label="networks", model="network")
```
CamLamb, line 13: Are you sure this works?
It looks like the tests are failing in CI because of this line

smanga24: I have done the changes

src/networks/migrations/0008_alter_network_network_type.py
```diff
@@ -0,0 +1,76 @@
...
+NEW_NETWORK_TYPE = "professional_networks_and_skills"
+
+
+def run_network_type_update(apps, old_network_type, new_network_type):
+    Network = apps.get_model("networks", "Network")
+    Revision = apps.get_model("wagtailcore", "Revision")
+    ContentType = apps.get_model("contenttypes", "ContentType")
+    content_type = ContentType.objects.get_or_create(
+        app_label="networks", model="network"
+    )[0]
```
CamLamb, line 14: We shouldn't be creating ContentTypes. This is something that is managed by Django and there is no operation in this migration that needs a content type that doesn't already exist.

smanga24: Test cases are failing because no ContentType available. we can create if no ContentType found or we can skip the migration.

PR #922, opened 2025-02-11
INTR-539 - Refactored initial_page_data and fixed missing values

david-okeke1337:
- [x] Tests are up to date

Moved logic for GTM initial page data out of template and into a templatetag.

Fixes and observations:

- Removed the trailing space from page_topics and user_professions
- Fixed missing "NA" when user_working_location is not set
- page_topics, page_tags and user_professions are all returned as space separated lists. This is not as much of an issue for user_professions as it returns the grade code (which contains no spaces), however page_topics and page_tags both can as seen in the example below.
- The above fields use an empty string to signify no data whereas the rest use "NA". Could this be consolidated for consistency?

None/Default values:

  - page_age_in_days: "NA"
  - page_type: "NA"
  - page_topics: ""
  - page_tags: ""
  - page_content_owner: "NA"
  - page_reactions: {}
  - user_professions: ""
  - user_grade: "NA"
  - user_working_location: "NA"

image

image


src/core/templates/includes/gtm_datalayer_info.html
```diff
@@ -23,5 +24,6 @@
     "user_working_location": "{{ request.user.profile.get_office_location_display|default_if_none:"NA" }}",
     "user_account_age_in_days": {{ request.user.profile.days_since_account_creation }}
 };
+initial_page_data_v2 = {% get_initial_page_data %}
```
CamLamb, line 27: Add below this `console.assert(initial_page_data == initial_page_data_v2)`
If we see an error in the console, that means that the new approach has an output different to the existing one and will help to be confident that the refactor is working as expected.

david-okeke1337: Just tested this, it works when I remove the fixes I made, looking at the existing logic for user_working_location it should return "NA" when empty however it currently returns an empty string, I also removed the trailing space from the concatenation for page_topics and user_professions (see screenshot above). Let me know if that makes sense or if I should leave the spaces + "NA" fix

david-okeke1337: Added comments to PR

src/core/templatetags/gtm_datalayer_info.py
```diff
@@ -0,0 +1,85 @@
...
+    for i, (k, v) in enumerate(context["FEATURE_FLAGS"].items()):
+        initial_page_data[f"feature_flag_{i+1}"] = f"{k}: {str(v).lower()}"
+
+    # Page Data
+    if "page" in context:
+        page = context["page"]
+        if isinstance(page, Page):
+            initial_page_data["page_age_in_days"] = get_page_age_in_days(page)
+            initial_page_data["page_type"] = get_page_type(page)
+            initial_page_data["page_topics"] = " ".join(
```
CamLamb, line 28: It's interesting that this was ever a "space separated" list, since topics can have spaces in their names 😆
Can you go through the output of `get_initial_page_data` and document this oddity along with any others so that we can start a chat with the analytics team and see if they'd be ok with us making this a "comma separated" list

src/core/templatetags/gtm_datalayer_info.py
```diff
@@ -0,0 +1,85 @@
...
+    )
+
+    return mark_safe(json.dumps(initial_page_data))  # noqa S308
+
+
+def get_page_age_in_days(page):
+    page_age_in_days = page.days_since_last_published
+    if page_age_in_days is not None:
+        return page_age_in_days
+    return "NA"
```
CamLamb, line 69:
```suggestion
def get_page_age_in_days(page):
    return page.days_since_last_published or "NA"
```

src/core/templatetags/gtm_datalayer_info.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def get_page_type(page):
+    content_type = ContentType.objects.get_for_model(page)
+    if content_type and hasattr(content_type, "name"):
+        return content_type.name
+    return "NA"
```
CamLamb, line 76: If `self.content_type.name` works in the template, does it not work `page.content_type.name` in python?

I would imagine this could look something like this (completely untested)

```suggestion
def get_page_type(page):
    if content_type := getattr(page, "content_type"):
        return content_type.name
    return "NA"
```

src/core/templatetags/gtm_datalayer_info.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def get_page_tags(page):
+    if hasattr(page, "tagged_items"):
+        page_tags = " ".join(
+            tagged_item.tag.name
+            for tagged_item in page.tagged_items.select_related("tag").all()
+        )
+    return page_tags if page_tags else ""
```
CamLamb, line 85:
```suggestion
def get_page_tags(page):
    page_tags = ""
    if hasattr(page, "tagged_items"):
        page_tags = " ".join(
            tagged_item.tag.name
            for tagged_item in page.tagged_items.select_related("tag").all()
        )
    return page_tags
```

CamLamb: OR even better:
```suggestion
def get_page_tags(page):
    if not hasattr(page, "tagged_items"):
        return ""

    page_tags = " ".join(
        tagged_item.tag.name
        for tagged_item in page.tagged_items.select_related("tag").all()
    )
    return page_tags if page_tags else ""
```

src/core/templatetags/gtm_datalayer_info.py
```diff
@@ -0,0 +1,85 @@
...
+    # User Data
+    initial_page_data["user_profile_slug"] = str(request.user.profile.slug)
+
+    for i, role in enumerate(request.user.profile.roles.all()):
+        initial_page_data[f"user_team_slug_{i+1}"] = role.team.name.lower()
+        initial_page_data[f"user_job_title_{i+1}"] = role.job_title
+
+    initial_page_data["user_professions"] = " ".join(
+        profession.code for profession in request.user.profile.professions.all()
+    )
```
CamLamb, line 48: Another space separated list

src/core/templatetags/gtm_datalayer_info.py
```diff
@@ -0,0 +1,85 @@
...
+    initial_page_data["user_grade"] = (
+        request.user.profile.grade.code if request.user.profile.grade else "NA"
```
CamLamb, line 50: With all of these `else "NA"` lines, it makes me think we should just create an initial dictionary:
```
initial_page_data = {
    ...
    "user_grade": "NA",
    ...
}
...
if request.user.profile.grade:
    initial_page_data["user_grade"] = request.user.profile.grade.code
...
It would get rid of all of the inline IFs 😄

File: src/core/templatetags/gtm_datalayer_info.py [hunk 8]
```diff
@@ -0,0 +1,85 @@
...
+    )
+    initial_page_data["user_is_line_manager"] = request.user.profile.is_line_manager
+    initial_page_data["user_working_location"] = (
+        request.user.profile.get_office_location_display()
+        if request.user.profile.get_office_location_display()
+        else "NA"
+    )
```
CamLamb, line 57: Creating the initial dictionary would mean that this gets simplified nicely too where we don't have to call the function twice:
```
if location := request.user.profile.get_office_location_display():
    initial_page_data["user_working_location"] = location
```

src/core/templatetags/gtm_datalayer_info.py
david-okeke1337: I've noted them down in the ticket/PR. I'm assuming we'll leave it as is until speaking with the analytics team?

src/core/templates/includes/gtm_datalayer_info.html
```diff
@@ -1,27 +1,6 @@
...
-    "user_team_slug_{{ forloop.counter }}": "{{ role.team.slug }}",
-    "user_job_title_{{ forloop.counter }}": "{{ role.job_title }}",
-    {% endfor %}
-    "user_professions": "{% for profession in request.user.profile.professions.all %}{{ profession.code }} {% endfor %}",
-    "user_grade": {% if request.user.profile.grade %}"{{ request.user.profile.grade.code }}"{% else %}"NA"{% endif %},
-    "user_is_line_manager": {{ request.user.profile.is_line_manager|lower }},
-    "user_working_location": "{{ request.user.profile.get_office_location_display|default_if_none:"NA" }}",
-    "user_account_age_in_days": {{ request.user.profile.days_since_account_creation }}
-};
+initial_page_data = {% get_initial_page_data %}
```
CamLamb, line 4: I hate to be a pain, but can we put this behind a feature flag so that no change happens at first, and we can toggle it on for staging and get the analytics team to test before we then enable it everywhere 😄

david-okeke1337: Sure, good idea!

CamLamb (approved):
Just one comment to address, then this is good to merge!

PR #928, opened 2025-02-17
INTR-541 Useful links sidebar section

nicopicchio:
- [ ] Tests are up to date

image


src/networks/models.py
```diff
@@ -158,6 +162,9 @@ class NetworkTypes(models.TextChoices):
         FieldPanel("is_peoplefinder_network"),
         FieldPanel("peoplefinder_network"),
     ]
+
+    promote_panels = NetworksHome.promote_panels
```
CamLamb, line 166: The reason this pattern is uses is to keep consistency in the inheritance tree, so you would only expect to see a class that the model inherits from being referenced.
`NetworksHome` isn't in the inheritance tree of the `Network` page, so this should just redefine the list like so.

```suggestion
    promote_panels = [
        FieldPanel("useful_links"),
    ]
```

I do like that you're thinking about preventing code duplication!

src/dw_design_system/dwds/icons/arrow-blue-bg.html
CamLamb, line 1: Can you add this icon to the list of icons that show on the page at `dwds/icons/`

src/dw_design_system/dwds/elements/useful_link.html
```diff
@@ -0,0 +1,9 @@
+<div class="content-stack medium border-top">
```
CamLamb, line 1: `border-top` is a behaviour, however it is defined so that it will be available everywhere.
We can give this element a specific class,`dwds-useful-links` and the styling for the top border can go on there

src/dw_design_system/dwds/elements/useful_link.html
CamLamb, line 1: The element is called `useful_link` however it contains many links.
Let's rename this to match: `useful_links.html`

src/dw_design_system/dwds/elements/content_icon.scss
```diff
@@ -38,7 +50,7 @@
 
     &.medium-gap {
         > .content-icon {
-            margin-inline-end: var(--s-3);
+            margin-inline-end: var(--s-2);
```
CamLamb, line 53: Changing styles that already exists will have side effects on existing components.
Have you checked the other usages of `medium-gap` and verified that the change doesn't look odd/have any undesirable changes?

src/dw_design_system/dwds/elements/useful_link.html
```diff
@@ -0,0 +1,9 @@
...
+    <h3 style="margin-top: var(--s-1)">{{ title }}</h3>
+    {% for link in useful_links %}
+        <div class="content-with-icon medium-gap top-alignment">
+            {% include "dwds/icons/arrow-blue-bg.html" with medium=True %}
+            <a href="{{ link.page }}" class="bold no-underline">{{ link.title }}</a>
```
CamLamb, line 6: Instead of a bold class, we can add the styling so that `a` tags inside this component are bold like so:
`.dwds-useful-links a`

CamLamb (commented):
Looking good!
Just some general naming and code tidy up!

src/dw_design_system/dwds/elements/content_icon.scss
nicopicchio: `medium-gap` is a new class I've created and it's not in use by anything else apart from `useful_links.html

CamLamb: Screenshot 2025-02-18 at 12 10 43
Just did a find on the `main` branch and I can see it being used elsewhere

src/dw_design_system/dwds/elements/useful_links.html
```diff
@@ -0,0 +1,9 @@
+<div class="dwds-useful-links content-stack medium ">
+    <h3 style="margin-top: var(--s-1)">{{ title }}</h3>
```
CamLamb, line 2: No inline styles, if we need to adjust the margin top, do it in the `dwds-useful-links` styles

CamLamb (approved):
Approved once comment is addressed

PR #953, opened 2025-03-04
INTR-569 Network group page

nicopicchio:
- [ ] Tests are up to date

image


src/networks/models.py
```diff
@@ -177,6 +177,11 @@ class NetworkTypes(models.TextChoices):
         ),
     ] + ContentOwnerMixin.indexed_fields
 
+    def get_template(self, request, *args, **kwargs):
+        if Network.objects.live().public().child_of(self).exists():
```
CamLamb, line 181: This logic should only run if the networks hub flag is active, otherwise it should just behave as it currently does

CamLamb (approved):

PR #955, opened 2025-03-06
INTR-341 Fix search breaking during deployment

CamLamb:
The search currently breaks because we have atomic rebuild enabled and we do 2 `update_index` calls, one with `--schema-only` and one without.

How it breaks:
- The current search index has documents
- The first `update_index` call with `--schema-only` builds a new index with just the schema changes and then changes the pointer for the search backend to use this new index (This index has NO documents)
- The second `update_index` call populates the documents and then changes the pointer over when done

How to test this locally:
- Start your local environment on the `main` branch
- Run `./manage.py update_index_async`
- Navigate to some search results (ideally with people in the response)
- Re run the `update_index_async` command and keep refreshing the search results page
  - If you are (un)lucky you will catch the scenario where some of the indexes have completed and some haven't, resulting in no/partial search results showing
- To test the fix, simply checkout this branch and rerun the above steps 😄 

- [ ] Tests are up to date


marcelkornblum (approved):
I don't remember why we did this in the first place but it seems very much the right change to make :D

PR #961, opened 2025-03-07
INTR-602 - Create the author component

david-okeke1337:
- [x] Tests are up to date

Testing:

The component should still be styled correctly regardless of whether profile_image_url is passed or not. (link/image)
[/dwds/components/#author](http://localhost:8000/dwds/components/#author)
image
image

### DIFF
--- a/src/dw_design_system/dwds/components/author.scss
+++ b/src/dw_design_system/dwds/components/author.scss
@@ -1,0 +1,20 @@
+.dwds-author {
+    &.content-sidebar {
+        --sidebar-gap: 0;
+        --sidebar-size: 60px;
+        --sidebar-min-size: 50%;
+        align-items: center;
+
+        > div:first-child {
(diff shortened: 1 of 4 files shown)


src/dw_design_system/dwds/components/author.scss
```diff
@@ -0,0 +1,20 @@
...
+        --sidebar-min-size: 50%;
+        align-items: center;
+
+        > div:first-child {
+            margin: var(--s-4);
+            text-align: center;
+            display: flex;
+            justify-content: center;
+
+            a:first-child {
```
CamLamb, line 14: There's only one `a` tag below the div so we can just say `> a` which will target only the `a` tags DIRECTLY below the first `div` under the `.content-sidebar`

```suggestion
            > a {
```

CamLamb (approved):

PR #974, opened 2025-03-18
INTR-599 - Edit comment

david-okeke1337:
- [ ] Tests are up to date


src/interactions/views.py
```diff
@@ -77,6 +84,52 @@ def react_to_page(request, *args, pk, **kwargs):
 
 
 @require_http_methods(["POST"])
+def edit_comment(request, *args, comment_id, **kwargs):
```
CamLamb, line 87: if you aren't using the `args` or `kwargs` but still want to set `comment_id` as a keyword argument, do can change the function like so:
```suggestion
def edit_comment(request, *, comment_id):
```

src/interactions/views.py
```diff
@@ -77,6 +84,52 @@ def react_to_page(request, *args, pk, **kwargs):
...
+        comment = comments_service.comment_to_dict(
+            get_object_or_404(Comment, id=comment_id)
+        )
+
+        return HttpResponse(comment["message"], content_type="text/html")
+
+    return HttpResponse(status=400)
+
+
+def edit_comment_form(request, *args, comment_id, **kwargs):
```
CamLamb, line 107: Similar here
```suggestion
def edit_comment_form(request, *, comment_id):
```

CamLamb: Also worth adding `@require_http_methods(["GET"])`

src/interactions/views.py
```diff
@@ -77,6 +84,52 @@ def react_to_page(request, *args, pk, **kwargs):
...
+    return TemplateResponse(
+        request,
+        "interactions/edit_comment_form.html",
+        context={
+            "comment": comment,
+        },
+    )
+
+
 def react_to_comment(request, *args, pk, **kwargs):
```
CamLamb, line 133: Just spotted this one, please could you make the others match if we aren't using the `args` or `kwargs`?

src/interactions/views.py
```diff
@@ -77,7 +84,114 @@ def react_to_page(request, *args, pk, **kwargs):
...
+        request,
+        "interactions/edit_comment_form.html",
+        context={
+            "comment": comment,
+        },
+    )
+
+
+# TODO: wip
+def toggle_edit_comment(request, *, comment_id, editing):
```
CamLamb, line 154: This view confuses me a bit. I think we should have the following views:
- Get comment (returns the comment template for a given comment)
- Get edit comment (returns the comment edit form for a given comment)
- Edit comment (the view that the comment edit form gets submitted to)

This toggle one is a little confusing as it knows a little too much about the user's current behaviour `editing`. This should be instead understood by the frontend and the view switched accordingly.
(I might be missing why this has been implemented this way though so happy to chat about this further)

src/interactions/templates/interactions/edit_comment_input.html
```diff
@@ -0,0 +1,11 @@
+{% load comments %}
+{% user_can_edit_comment request.user comment.id as can_edit %}
+{% if can_edit %}
+    {% if comment.edit_comment_form_url %}
+        <button class="dwds-button dwds-button--link"
+                id="edit-comment-{{ comment.id }}-form-show"
```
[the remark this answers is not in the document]
david-okeke1337: Good spot! Was from when there were 2 reply buttons

src/interactions/templatetags/comments.py
```diff
@@ -33,14 +29,46 @@ def page_comments(context, page: Page):
         comments.append(comment)
     return {
         "user": context["user"],
-        "comment_count": get_page_comment_count(page),
+        "comment_count": comments_service.get_page_comment_count(page),
         "comments": comments,
         "comment_form": CommentForm(),
         "comment_form_url": comment_form_submission_url,
         "request": context["request"],
+        "page": page,
```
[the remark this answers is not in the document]
david-okeke1337: Removed!

src/dw_design_system/dwds/components/comment.html
```diff
@@ -1,20 +1,26 @@
...
             {% reactions_for_comment user comment.id %}
+
             {% if comment.reply_form %}
                 <span class="separator"></span>
                 <button class="dwds-toggle-visibility dwds-button dwds-button--link"
                         data-target-identifier="comment-{{ comment.id }}-reply-form"
                         data-target-state="hidden">Reply</button>
             {% endif %}
             {% include "news/includes/delete_comment.html" with comment=comment %}
+            {% user_can_edit_comment request.user comment.id as can_edit %}
```
CamLamb, line 19: This check is done in the `edit_comment_input` template, if we move the separator into that template we can get rid of the check from this template

CamLamb (approved):
Approved with one comment to address

PR #979, opened 2025-03-20
Update the comment styling

CamLamb:
Had a chat with Cheryl and made some tweaks based on the designs and discussion.


nicopicchio (approved):
looking good!

PR #980, opened 2025-03-24
External useful links

CamLamb:
Add the ability to link externally for "Useful Links" and add external link text to all links on the site (behind a flag so we can test and configure before rolling out)

Examples:
Screenshot 2025-03-24 at 11 30 51
Screenshot 2025-03-24 at 11 30 58


david-okeke1337 (approved):
Looks good! You may want to add it to the components page as I mentioned in a comment

PR #993, opened 2025-03-28
Tests for comments

nicopicchio:
- [ ] Tests are up to date


src/interactions/test/views/test_comments_views.py
CamLamb, line 1: This file name doesn't need to say `views` in it as the directory it is in is the `views` directory.

This file could just be `test_comments.py` as we know that it is testing the views from the directory it is in

CamLamb: Can we also add tests for the other `comments` views here:
- `comment_on_page`
- `edit_comment`
- `edit_comment_form`
- `get_comment`

CamLamb (approved):

PR #1003, opened 2025-04-03
Add an initial lighthouse action for testing

CamLamb:


CamLamb:
Oddly, the URL doesn't seem to make a difference to the report that gets generated, perhaps there's a redirect causing some issues

PR #1005, opened 2025-04-04
Add Django Cotton and test it with the Modal

CamLamb:


CamLamb:
Some thoughts about implementation, should we name our components `dwds-modal.html` so that the usage becomes `<c-dwds-modal ...>`

nicopicchio (approved):

PR #1014, opened 2025-04-08
INTR-643 -  Comments tests

david-okeke1337:
- [ ] Tests are up to date

### DIFF
--- a/src/interactions/services/comments.py
+++ b/src/interactions/services/comments.py
@@ -2,6 +2,7 @@ from datetime import datetime
 
 from django.db import models
 from django.db.models import QuerySet
+from django.http import HttpRequest
 from django.shortcuts import get_object_or_404
 from django.template.response import TemplateResponse
 from django.urls import reverse
@@ -144,12 +145,19 @@ def hide_comment(comment: Comment) -> None:
 def can_hide_comment(user: User, comment: Comment | int) -> bool:
     if isinstance(comment, int):
         comment = Comment.objects.get(id=comment)
+
+    assert isinstance(comment, Comment)
     return user == comment.author
 
 
 def can_edit_comment(user: User, comment: Comment | int) -> bool:
     if isinstance(comment, int):
-        comment = Comment.objects.get(id=comment)
+        try:
+            comment = Comment.objects.get(id=comment)
+        except Comment.DoesNotExist:
+            return False
+
+    assert isinstance(comment, Comment)
     return user == comment.author
 
 
@@ -160,10 +168,11 @@ def can_reply_comment(user: User, comment: Comment | int) -> bool:
         except Comment.DoesNotExist:
             return False
 
-    return bool(not comment.parent)
+    assert isinstance(comment, Comment)
+    return not bool(getattr(comment, "parent", False))
 
 
-def get_page_comments_response(request, page: Page) -> TemplateResponse:
+def get_page_comments_response(request: HttpRequest, page: Page) -> TemplateResponse:
     comments = [
         comments_service.comment_to_dict(page_comment)
         for page_comment in comments_service.get_page_comments(page)
--- a/src/interactions/test/conftest.py
+++ b/src/interactions/test/conftest.py
@@ -32,6 +32,15 @@ def news_page():
 
 
 @pytest.fixture
+def news_page2():
+    return NewsPage.objects.create(
+        title="News 2",
+        depth=1,
+        path="page3",
+    )
+
+
+@pytest.fixture
 def about_page():
     return AboutUs.objects.create(
         title="About Us",
@@ -48,8 +57,16 @@ def create_page_reaction(user, news_page):
 
 
 @pytest.fixture
-def comment(user, news_page):
-    return Comment.objects.create(author=user, page=news_page, content="a comment")
+def comment(comment_factory):
+    return comment_factory()
+
+# TODO: switch to kwargs for comment_factory and add back default values 
+@pytest.fixture
+def comment_factory(user, news_page):
+    def create_comment(author=user, page=news_page, content="a comment"):
+        return Comment.objects.create(author=author, page=page, content=content)
+
+    return create_comment
 
 
 @pytest.fixture
--- a/src/interactions/test/conftest.py
+++ b/src/interactions/test/conftest.py
@@ -2,9 +2,10 @@ import pytest
 from django.contrib.auth import get_user_model
 
 from about_us.models import AboutUs
-from news.models import NewsPage, Comment
+from news.models import NewsPage
 
 from interactions.models import PageReaction, ReactionType, CommentReaction
+from news.factories import CommentFactory
 
 
 @pytest.fixture
@@ -32,15 +33,6 @@ def news_page():
 
 
 @pytest.fixture
-def news_page2():
-    return NewsPage.objects.create(
-        title="News 2",
-        depth=1,
-        path="page3",
-    )
-
-
-@pytest.fixture
 def about_page():
     return AboutUs.objects.create(
         title="About Us",
@@ -48,25 +40,17 @@ def about_page():
         path="page1",
     )
 
+# TODO: Update test_comment_reactions.py usage
+@pytest.fixture
+def comment():
+    return CommentFactory.create()
+
 
 @pytest.fixture
 def create_page_reaction(user, news_page):
     return PageReaction.objects.create(
         user=user, page=news_page, type=ReactionType.LIKE
     )
-
-
-@pytest.fixture
-def comment(comment_factory):
-    return comment_factory()
-
-# TODO: switch to kwargs for comment_factory and add back default values 
-@pytest.fixture
-def comment_factory(user, news_page):
-    def create_comment(author=user, page=news_page, content="a comment"):
-        return Comment.objects.create(author=author, page=page, content=content)
-
-    return create_comment
 
 
 @pytest.fixture
--- a/src/interactions/test/services/test_comments.py
+++ b/src/interactions/test/services/test_comments.py
@@ -1,5 +1,6 @@
 import pytest
 from interactions.services import comments as comments_service
+from news.models import Comment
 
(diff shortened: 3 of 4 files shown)


src/interactions/services/comments.py
```diff
@@ -144,12 +145,19 @@ def hide_comment(comment: Comment) -> None:
...
     return user == comment.author
 
 
 def can_edit_comment(user: User, comment: Comment | int) -> bool:
     if isinstance(comment, int):
-        comment = Comment.objects.get(id=comment)
+        try:
+            comment = Comment.objects.get(id=comment)
+        except Comment.DoesNotExist:
+            return False
```
CamLamb, line 158: We do a try/except here, but we don't in `can_hide_comment`
We should be consistent in our approach.

Personally, I think that if we are given an integer and it isn't a valid comment ID, then we should let the exception bubble up (no try/except).

src/interactions/services/comments.py
```diff
@@ -160,10 +168,11 @@ def can_reply_comment(user: User, comment: Comment | int) -> bool:
         except Comment.DoesNotExist:
             return False
```
CamLamb, line 169: Similarly, we should be consistent here with the approach we decide.

src/interactions/test/services/test_comments.py
```diff
@@ -1,58 +1,258 @@
...
+        news_page, user, reply_content, comment.pk
+    )
+
+    # The comment reply is created succesfully
+    assert isinstance(reply, Comment)
+    assert reply.content == reply_content
+    assert reply.parent == comment
+
+
+def test_get_page_comments(news_page):
```
CamLamb, line 84: Please can you add some comments with `is_visble=False` to these to make sure they don't show up?

src/interactions/test/services/test_comments.py
```diff
@@ -1,58 +1,258 @@
...
+
+    for news_comment in news_comments:
+        # The comments are returned for the requested page
+        assert news_comment in news_page_comments
+
+    # Comments on other pages are not returned
+    assert comment not in news_page_comments
+
+
+def test_get_page_comment_count(news_page):
```
CamLamb, line 97: Some `is_visible=False` tests here would be great also!

src/interactions/test/services/test_comments.py
```diff
@@ -1,58 +1,258 @@
...
+    comments = CommentFactory.create_batch(3, page=news_page)
+
+    # The comment count is returned for the requested page
+    assert comments_service.get_page_comment_count(news_page) == len(comments)
+
+
+def test_get_comment_replies():
```
CamLamb, line 104: Also ome `is_visible=False` tests here 😄

src/interactions/test/services/test_comments.py
```diff
@@ -1,58 +1,258 @@
...
+    comment = CommentFactory.create()
+    replies = CommentFactory.create_batch(3, page=comment.page, parent=comment)
+    comment_replies = comments_service.get_comment_replies(comment)
+
+    for reply in replies:
+        # The replies are returned for the requested comment
+        assert reply in comment_replies
+
+
+def test_get_comment_reply_count():
```
CamLamb, line 114: Some `is_visible=False` comments here

src/news/factories.py
```diff
@@ -10,3 +11,19 @@ class NewsPageFactory(PageWithTopicsFactory, wagtail_factories.PageFactory):
 
     class Meta:
         model = models.NewsPage
+
+
+class CommentFactory(factory.django.DjangoModelFactory):
+    author = factory.Sequence(
+        lambda n: get_user_model().objects.create(username=f"user_{n+1}")
```
CamLamb, line 18: I think there is a `UserFactory` and a `PersonFactory`
So you could use both of those to create a "whole" user and assign them as the author

src/interactions/test/views/test_comments.py
```diff
@@ -1,29 +1,163 @@
...
+    comment = CommentFactory()
+    client = Client()
+    client.force_login(user)
+
+    url = lambda comment_id: reverse("interactions:get-comment", args=[comment_id])
+
+    # Test view when a valid comment is passed
+    response = client.get(url(comment.id))
+    assert response.status_code == 200
+    assert comment.content in response.content.decode()
```
CamLamb, line 110:
```suggestion
    assert comment.content in response.content.decode()
    assert response.context["comment"]["id"] == comment.id
```

src/interactions/test/views/test_comments.py
```diff
@@ -1,29 +1,163 @@
...
+    response = client.get(url(123456))
+    assert response.status_code == 404
+
+
+def test_get_page_comments_view(user, news_page):
+    comments = CommentFactory.create_batch(3, page=news_page)
+    client = Client()
+    client.force_login(user)
+
+    url = lambda page_id: reverse("interactions:get-page-comments", args=[page_id])
```
CamLamb, line 122: Same as above

src/interactions/test/views/test_comments.py
```diff
@@ -1,29 +1,163 @@
...
+    for comment in comments:
+        assert f"comment-{comment.id}" in response.content.decode()
+        assert comment.content in response.content.decode()
+
+    # Test view when an invalid page is passed
+    response = client.get(url(123456))
+    assert response.status_code == 404
+
+
+def test_reply_to_comment_view(mocker, user):
```
CamLamb, line 136: Add a check for the 404 scenario

src/interactions/test/views/test_comments.py
```diff
@@ -1,29 +1,163 @@
...
+
+    url = reverse("interactions:reply-comment", args=[comment.id])
+    comment_content = "a comment reply"
+
+    request = {
+        "path": url,
+        "data": {
+            "comment": comment_content,
+        },
+    }
```
CamLamb, line 149: Same as previous

src/interactions/test/views/test_comments.py
```diff
@@ -1,29 +1,163 @@
...
+
+    # Test view when the user IS authorised to reply to the comment
+    mocker.patch("interactions.services.comments.can_reply_comment", return_value=True)
```
CamLamb, line 152: Decorate the test with this patch

src/interactions/test/views/test_comments.py
```diff
@@ -1,29 +1,163 @@
...
+    response = client.post(**request)
+    reply = Comment.objects.get(parent=comment)
+    assert reply.content == comment_content
+    assert response.status_code == 200
+    assert comment_content in response.content.decode()
```
CamLamb, line 157:
```suggestion
    assert comment_content in response.content.decode()
    assert response.context["comment"]["id"] == comment.id
```

src/core/test/conftest.py
CamLamb, line 1: I wonder if we make this a project wide change?
maybe this file should be moved to `src/tests/conftest.py` and set the `scope` to `session`

src/interactions/test/conftest.py
```diff
@@ -1,34 +1,23 @@
...
 
 
 @pytest.fixture
 def news_page():
-    return NewsPage.objects.create(
-        title="News",
-        depth=1,
-        path="page2",
-    )
+    news_home_page = NewsHome.objects.first()
```
CamLamb, line 19: `first` returns `NewsHome | None`
This means that there is a chance you create a `NewsPage` with no parent. Probably worth changing this to `get` so that it fails hard if there is no `NewsHome`

src/interactions/test/services/test_comments.py
```diff
@@ -1,58 +1,310 @@
...
+    replies = CommentFactory.create_batch(3, page=comment.page, parent=comment)
+
+    # Hidden replies are not retrieved
+    CommentFactory.create_batch(3, page=comment.page, parent=comment, is_visible=False)
+
+    comment_replies = comments_service.get_comment_replies(comment)
+
+    for reply in replies:
+        # The replies are returned for the requested comment
+        assert reply in comment_replies
```
CamLamb, line 128: This just tests that the reply you created is in the response, it doesn't test if the hidden comments aren't

CamLamb, line 128:
```suggestion
    # Hidden replies are not retrieved
    hidden_replies = CommentFactory.create_batch(3, page=comment.page, parent=comment, is_visible=False)

    comment_replies = comments_service.get_comment_replies(comment)

    for comment_reply in comment_replies:
        assert comment_reply in replies
        assert comment_reply not in hidden_replies
```

src/interactions/test/services/test_comments.py
```diff
@@ -1,58 +1,310 @@
...
+    # No replies are returned for the new comment
+    assert comments_service.get_comment_reply_count(comment) == 0
+
+    replies = CommentFactory.create_batch(3, page=comment.page, parent=comment)
+
+    # Hidden replies are not counted
+    CommentFactory.create_batch(3, page=comment.page, parent=comment, is_visible=False)
+
+    # The correct number of replies are returned for the requested comment
+    assert comments_service.get_comment_reply_count(comment) == len(replies)
```
CamLamb, line 142: We should create different amounts of hidden vs visible comments, otherwise we could break the `get_comment_reply_count` so it counts only hidden comments and the tests won't catch it

src/interactions/test/services/test_comments.py
```diff
@@ -1,58 +1,311 @@
...
+
+    # Comments on other pages are not returned
+    assert comment not in news_page_comments
+
+
+def test_get_page_comment_count(news_page):
+    comments = CommentFactory.create_batch(3, page=news_page)
+
+    # Hidden comments are not included in the comment count
+    CommentFactory.create_batch(3, page=news_page, is_visible=False)
```
CamLamb, line 111: can you create a batch with a different size, otherwise we don't know that the 3 we get back in the test is because of the hidden or visible ones :)

CamLamb (approved):

PR #1015, opened 2025-04-09
IDPF-393 Add pf_discover flag

alsyx:


CamLamb (approved):
There's not really much else to do here tbh.
Usually we add the flag in the same PR as the functionality that needs it since it's such a tiny change

PR #1033, opened 2025-04-15
Make some styling changes for discussion

CamLamb:
- Give cards a sensible minimum width
  - This means that if a card is about to get too narrow, it will wrap onto the next line
  - [x] TODO: Fix the engagement component image when it is horizontal
- Make the sidebar as small as the minimum width a card a can be
  - This means that there is more room for the main page content
- Set a minimum width for the main page content (when there is a sidebar) to be the size of a card with some spacing surrounding it
  - This means that the contents of a page won't get too narrow that it cuts the contents of a card
- Fix the Ribbon border radius
- Remove the `width: fit-content` from `<a>` tags
  - [x] TODO: Fix links that go full width (in stacks)
- Fix event images so they aren't rounded incorrectly

**Before/After (large screen)**
Screenshot 2025-04-15 at 18 46 12

**Before/After (slightly smaller screen)**
Here we can see that the new changes won't allow the 3 cards to be really thin side-by-side-by-side, instead the last one wraps and all of the cards take up the gained room.
Screenshot 2025-04-15 at 18 47 28


david-okeke1337 (approved):

CamLamb:
After discussion with Cheryl, we decided to make the switcher behave as it should and go from horizontal to vertical stack instead of squishing the cards to be way too narrow

PR #1037, opened 2025-04-17
IDPF-399 Add filters to discover view

alsyx:
Added the filters for discover view. 
- adding cache so we don't have to make expensive db calls each time we apply the filters.
- refactoring where the query functions are.

- [ ] Tests are up to date


marcelkornblum (commented):
The whole approach of spreading this functionality across utils.py and filters.py feels like an anti-pattern to me; surely these should all be part of what's in our service?
having said that, the functionality actually in here seems OK i think

src/peoplefinder/utils.py
```diff
@@ -8,3 +16,59 @@ def get_uk_city_locations() -> list[tuple[str]]:
...
+    return [(network, network) for network in networks]
+
+
+def get_additional_roles() -> list[tuple[str]]:
+    additional_roles = (
+        AdditionalRole.objects.order_by("name")
+        .values_list("name", flat=True)
+        .distinct()
+    )
+    return [(role, role) for role in additional_roles]
```
CamLamb, line 74: So many database calls!
We might want to think about caching these functions

src/peoplefinder/filters.py
```diff
@@ -1,7 +1,16 @@
...
+from peoplefinder.utils import (
+    get_additional_roles,
+    get_grades,
+    get_key_skills,
+    get_learning_interests,
+    get_networks,
+    get_professions,
+    get_uk_buildings,
+    get_uk_city_locations,
+)
```
CamLamb, line 13: it feels like all of these belong in a service

src/peoplefinder/services/directory.py
```diff
@@ -16,3 +17,11 @@ def get_people(user: User) -> QuerySet[Person]:
         queryset = queryset.active_or_inactive_within(days=days)
 
     return queryset
+
+
+def get_people_with_filters(
+    *,
+    filter_options: list[tuple[str]],
```
CamLamb, line 24: Is this type correct?
It comes from `request.GET` and is fed into `DiscoverFilters.data`
My guess would be that it should be something like a dict (keys and values)

src/peoplefinder/services/reference.py
```diff
@@ -0,0 +1,99 @@
...
+    Profession,
+    UkStaffLocation,
+)
+
+
+def query_data(*, model: Any, field_name: str) -> list[tuple[str]]:
+    data = (
+        model.objects.order_by(field_name).values_list(field_name, flat=True).distinct()
+    )
+    return [(value, value) for value in data]
```
marcelkornblum, line 19: I'm not against this optimisation, buuuuutttt....

I think this arguably _does_ belong in a utils (in core?) and I think could do with more easily understood function and parameter naming. It's such a specific output format though maybe I'm wrong about that... (either way feels not part of the "what can I do with reference data" thing for services)

src/peoplefinder/services/reference.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def get_uk_city_locations() -> list[tuple[str]]:
+    cache_key = "peoplefinder.services.reference.get_uk_city_locations"
```
marcelkornblum, line 23: absolutely not a requirement, but give you've been trying to get optimisations going, you could make the whole thing into a factory

e.g.
```py
import inspect

def get_cached_ref_data(model, field_name):
    cache_key = f"{inspect.currentframe().f_back.f_code.co_name}.{inspect.currentframe().f_code.co_name}"
    if not (data := cache.get(key=cache_key)):
        data_list = (
            model.objects.order_by(field_name).values_list(field_name, flat=True).distinct()
        )
        data = [(value, value) for value in data_list]
        cache.set(key=cache_key, value=data, timeout=60 * 60)

    return data
    
def get_uk_city_locations():
    return get_cached_ref_data(model=UkStaffLocation, field_name="city")
```

(NB untested!!)

src/peoplefinder/templates/peoplefinder/includes/discover_filters.html
```diff
@@ -8,6 +7,32 @@ <h3>Sort and filter</h3>
                 {{ discover_filters.form.city }}
             </div>
             <div class="content-stack">
+                <label for="id_building_name">Filter by building name</label>
+                {{ discover_filters.form.building_name }}
+            </div>
+            <div class="content-stack">
+                <label for="id_grade">Filter by grade</label>
+                {{ discover_filters.form.grade }}
+            </div>
```
marcelkornblum, line 16: absolutely fine as is, but also clearly could be optimised with a loop if you're wanting to do that

marcelkornblum (approved):
Good as it is, but I wouldn't mind at least a little better name on that helper function, if not a full blown journey into optimisation :)

src/peoplefinder/services/reference.py
marcelkornblum: (to be a factory it ought to return a function but close enough - this is almost what Lisp would call a Macro :)

CamLamb: I love the idea, but that cache key hurts my eyes!

It could just be:
`cache_key = f"{model.__module__.__name__}.{model.__name__}"

marcelkornblum: i knew there'd be a nicer way but I couldn't work it out and SO kept insisting on inspect every time I tried looking 😅

marcelkornblum: i prefer the decorator approach anyway

src/peoplefinder/services/reference.py
alsyx: I added a comment to the function, moved it to `core.utils` and updated `model: models.Model` 
https://github.com/uktrade/digital-workspace-v2/pull/1037/files#diff-fe10ddddd5cbd1688cdcbcbc176c9fed764f8c50afa985851e98c5018fcadb44R104

PR #1039, opened 2025-04-22
IDPF-396 Add sort by first_name to discover view

alsyx:
Added first ordering option, sort by `first_name`.
Updated the `people_set` in the discover view to be in ascending order of `first_name` if no ordering option is selected.

- [ ] Tests are up to date


src/peoplefinder/views/directory.py
```diff
@@ -84,7 +84,7 @@ def discover(request: HttpRequest) -> HttpResponse | HttpResponseRedirect:
     if not flag_is_active(request, flags.PF_DISCOVER):
         return redirect("people-directory")
 
-    people_set = directory_service.get_people(request.user)
+    people_set = directory_service.get_people(request.user).order_by("first_name")
```
marcelkornblum, line 87: this looks hardcoded rather than wired into your filter, is that right?

marcelkornblum (commented):
I'm not sure putting this into the filters is the right way to do it; it looks a little hacky to me, but I may be wrong.

src/peoplefinder/views/directory.py
alsyx: yes this is not in the filter, i thought i would put it here as the default behaviour. I can look into creating a default filter for it instead.

marcelkornblum (approved):

CamLamb (approved):

PR #1040, opened 2025-04-23
IDPF-397 Build seniority, add grade sorting option

alsyx:
- Added sorting by grade, and the ability to apply multiple sorting options at the same time.
- Added ordering to meta class of `Person` model.
- Added a new field `ordering` to `Grade` model (higher the grade, higher the ordering value) and by default Grade is ordered by descending order of the ordering field, so we can get the highest grade first.
-- eg. Executive officer has ordering=3 and the scs_4 has ordering=11 and when we order them desc, we get scs_4 then Executive officer.

- [ ] Tests are up to date


src/peoplefinder/migrations/0126_insert_grade_ordering_data.py
```diff
@@ -0,0 +1,37 @@
...
+    Grade.objects.filter(code="scs_4").update(ordering=12)
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("peoplefinder", "0125_alter_grade_options_grade_ordering"),
+    ]
+
+    operations = [migrations.RunPython(insert_grade_ordering_data)]
```
CamLamb, line 37: Add a reverse

marcelkornblum (approved):

PR #1041, opened 2025-04-23
IDPF-392: PF discover page pagination

marcelkornblum:
Basic pagination implementation using DWDS component

### DIFF
--- a/src/peoplefinder/templates/peoplefinder/includes/discover_filters.html
+++ b/src/peoplefinder/templates/peoplefinder/includes/discover_filters.html
@@ -1,26 +1,18 @@
-<div class="dwds-content-item-card no-shadow">
-    <div class="content-item">
-        <div class="content-main large highlight">
-            <div class="content-body">
-                <form method="get" class="content-stack large">
-                    <h1>Discover DBT staff</h1>
-                    <hr>
-                    <div class="content-stack large">
-                        <div class="content-switcher">
-                            <div class="content-stack">
-                                <label for="id_city">Filter people by city</label>
-                                {{ discover_filters.form.city }}
-                            </div>
-                        </div>
-                    </div>
-                    <div class="dwds-button-group">
-                        <input class="dwds-button dwds-button--secondary dwds-button--inline"
-                               type="submit"
-                               value="Apply filters" />
-                        <a href="?">Reset filters</a>
-                    </div>
-                </form>
+
+<form method="get" class="content-stack large">
+    <h3>Sort and filter</h3>
+    <div class="content-stack large">
+        <div class="content-switcher">
+            <div class="content-stack">
+                <label for="id_city">Filter by city</label>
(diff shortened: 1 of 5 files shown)


src/peoplefinder/templates/peoplefinder/includes/discover_filters.html
```diff
@@ -1,26 +1,18 @@
-<div class="dwds-content-item-card no-shadow">
-    <div class="content-item">
-        <div class="content-main large highlight">
-            <div class="content-body">
```
marcelkornblum, line 4: I'd rather have undesigned elements look undesigned
also - this is emphatically NOT a content item...

src/peoplefinder/urls.py
```diff
@@ -41,7 +41,7 @@
 people_urlpatterns = [
     path("", PeopleHome.as_view(), name="people-home"),
     path("directory/", PeopleDirectory.as_view(), name="people-directory"),
-    path("discover/", discover, name="discover"),
+    path("discover/", discover, name="people-discover"),
```
marcelkornblum, line 44: just for consistency with other urls in this file

src/peoplefinder/views/directory.py
```diff
@@ -87,9 +88,19 @@ def discover(request: HttpRequest) -> HttpResponse | HttpResponseRedirect:
...
+    except paginator.EmptyPage:
+        return redirect(
+            "people-discover", page=1
+        )  # would be nice to have some sort of mnessage with this
+
     context = {
         "page_title": "Discover",
-        "people": people,
+        "pages": pages.page(page),
+        "people": pages.page(page).object_list,
```
alsyx, line 103: how are we going to apply the filters if we don't get people from `people = discover_filters.qs` ?

alsyx: ah sorry i just saw the pages use `people = discover_filters.qs`

src/peoplefinder/views/directory.py
```diff
@@ -87,9 +88,18 @@ def discover(request: HttpRequest) -> HttpResponse | HttpResponseRedirect:
...
     discover_filters = DiscoverFilters(request.GET, queryset=people_set)
     people = discover_filters.qs
+
+    pr = paginator.Paginator(people, per_page=30)
+    page: int = int(request.GET.get("page", default=1))
+    try:
+        paginator_page = pr.page(page)
+    except paginator.EmptyPage:
+        # would be nice to have some sort of mnessage with this
+        return redirect("people-discover")
```
CamLamb, line 98: What if page 1 is empty? This introduces the potential for an infinite redirect

marcelkornblum: you're right; i'll let it get handled by the 'no matching colleagues' message instead

alsyx (approved):

PR #1044, opened 2025-04-23
Quote styling

CamLamb:
Update the quote styling and behaviour after a discussion with Cheryl

### DIFF
--- a/src/dw_design_system/dwds/components/quote.scss
+++ b/src/dw_design_system/dwds/components/quote.scss
@@ -41,8 +41,14 @@
     }
 
     > .content-sidebar {
-        --sidebar-size: 160px;
+        --sidebar-size: 130px;
+        --sidebar-min-size: 200px;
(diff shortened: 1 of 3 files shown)


src/dw_design_system/dwds/components/quote.scss
```diff
@@ -41,8 +41,14 @@
     }
 
     > .content-sidebar {
-        --sidebar-size: 160px;
+        --sidebar-size: 130px;
+        --sidebar-min-size: 200px;
```
marcelkornblum, line 45: is this intentionally smaller than the size in the previous line?

marcelkornblum (approved):

src/dw_design_system/dwds/components/quote.scss
CamLamb: Yah, one of Cheryl's comments was to make the image smaller

PR #1049, opened 2025-04-25
Run `strip_tags` on text based fields before storing values in the DB

CamLamb:
TODO: WRITE


marcelkornblum:
I think it would be useful in the long run to have a PR with the canonical implementation distinct from the application across all the apps; easier to parse, understand and share around. Not sure if it's really worth the effort though.

CamLamb:
Tests are currently failing because we are now calling `full_clean` on save for models with fields that weren't initially built with this approach in place

PR #1055, opened 2025-04-29
IDPF-421 Optimise get_people query and fix ordering for discover page

alsyx:
Attempt to optimise the `get_people` query. Fixed the ordering issue, people with `grade=None` will appear at the end.

- [ ] Tests are up to date


src/peoplefinder/services/directory.py
```diff
@@ -13,6 +13,22 @@ def get_people(user: User) -> QuerySet[Person]:
...
+        )
+        .prefetch_related(
+            "roles__team",
+            "key_skills",
+            "workdays",
+            "learning_interests",
+            "networks",
+            "professions",
+            "additional_roles",
+        )
```
marcelkornblum, line 31: im not sure we need all this and itll actually slow the query down if we're not using it all. 
since for discovery we know which fields we're displaying (for now) let's only select/prefetcvh the ones we'll use in the UI; we can come up with a more complete / optimised way to do this down the line

0RWilliams (approved):

src/peoplefinder/models.py
```diff
@@ -56,7 +56,7 @@ class Meta:
             models.UniqueConstraint(fields=["code"], name="unique_grade_code"),
             models.UniqueConstraint(fields=["name"], name="unique_grade_name"),
         ]
-        ordering = ["-ordering"]
+        ordering = ["ordering"]
```
marcelkornblum, line 59: lol @ switching this round to "top is first" ordering :D

marcelkornblum (commented):
let's trim that prefetch/select as below pls

src/peoplefinder/models.py
alsyx: you win 🎉 :D

src/peoplefinder/services/directory.py
CamLamb: I think for `select_related`, the following options are likely not needed:
- `user`
- `manager`

And for `prefetch_related`, we likely only need to fetch objects that we present to the user. For example: we aren't showing any values from `key_skills`, so we probs don't need that

marcelkornblum: we dont display any of the `select_related` fields in the current discovery view i think, let's drop them all. might be nice in the `get_people_with_filters` to add the ones we need, at the point we know which filters we're using

marcelkornblum (approved):

PR #1058, opened 2025-04-30
Optimise discover page

marcelkornblum:
Use function caching and query optimisation to improve performance for the discover page, but also other things we can find. Specifically:

* extended the cache_for_one_hour decorator to be more flexible (but then didnt use it anywhere new)
* saved all the feature flag info to the session and made a drop-in replacement for the main waffle accessor, using the session data (saving several DB calls on each page)
* the BIGGEST improvement for the discover page was simply dropping the use of the `get_annotated` queryset method; on its own this reduced the main run tom of the main query on that page by over a factor of 100
* Added more prefetch and select related items to the discover service
* went down rabbit holes adding extra related queries for filters, but these werent helpful
* went down rabbit holes investigating optimisations for the wagtailmenus dep but these involve tearing it to pieces
* went down rabbit holes seeing if clever stuff could be done at auth return level to have a user that wouldn't need extra DB calls when e.g. populating the GTM layer but this didn't speed anything up in a way that felt worthwhile


CamLamb (approved):

src/core/filters.py
```diff
@@ -3,4 +3,7 @@
 
 class FilterSet(django_filters.FilterSet):
     def has_filters_applied(self) -> bool:
-        return any(self.form.cleaned_data.values())
+        return any(self.form.changed_data)
+
+    def get_applied_filters(self):
```
CamLamb, line 8: Is this used?

src/peoplefinder/models.py
```diff
@@ -894,6 +894,16 @@ def get_roles_display(self) -> Optional[str]:
...
 
+    def roles_str(self):
+        output = ""
+        for role in self.roles.select_related("team").all():
+            if role.job_title:
+                output += f"{role.job_title} in"
+            else:
+                output += "Member of"
+            output += f" {role.team.name}"
+        return output
```
CamLamb, line 905: How does this look if a user has multiple roles?

CamLamb (approved):

src/core/filters.py
marcelkornblum: no, i put it in while fiddling, thought it might be useful and then backed out of using it in the end

src/peoplefinder/models.py
marcelkornblum: it doesn't work nicely 🤦

src/peoplefinder/services/directory.py
```diff
@@ -12,12 +12,18 @@ def get_people(user: User) -> QuerySet[Person]:
     """
     queryset = (
         Person.objects.all()
-        .get_annotated()
```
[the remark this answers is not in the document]
marcelkornblum: that feels possible, but really complex and perhaps wouldn't get too much extra performance. Feels perhaps worth doing but maybe an enhancement?

alsyx (approved):

PR #1059, opened 2025-05-01
IDPF-400 Is active filter

0RWilliams:
Added the is_active filter to the discovery page.
Limited the filters visibility to users that have can_view_inactive_profiles role.
Refactored template to avoid using multiple if's & hardcoded strings, allows for better custom behaviour. (Still a String compare rather than the object one i wanted. 😢 )

- [ ] Tests are up to date


alsyx (approved):

src/peoplefinder/filters.py
```diff
@@ -89,6 +99,18 @@ class DiscoverFilters(FilterSet):
...
+
+    def apply_active_users(self, queryset, name, value) -> QuerySet[Person]:
+        match value:
+            case "active":
+                return queryset.filter(is_active=True)
+            case "inactive":
+                return queryset.filter(is_active=False)
+            case "_":
+                return queryset
+
```
CamLamb, line 113: Does this need to be a custom method filter?
I think setting `is_active` as the field and setting the labels for the choices would achieve this with less code

CamLamb: Yeah, thats clean!

PR #1063, opened 2025-05-01
IDPF-412 Add is_civil servant filter

alsyx:
Excludes non-graded contractors, not sure if we need to exclude any other grade.

- [ ] Tests are up to date


src/peoplefinder/filters.py
```diff
@@ -33,7 +33,13 @@
 
 
 class DiscoverFilters(FilterSet):
-
+    is_civil_servant = django_filters.ChoiceFilter(
```
0RWilliams, line 36: I'd argue this is a checkbox rather than ChoiceFilter, but it should work either way.
I played around with checkboxes for is_active, but went with Choices as i wanted to be explicit with both options. i.e. query only active, query only inactive's and apply no filter.

marcelkornblum: I agree - and getting a few checkboxes in will hopefully stimulate the discussion i reckon

src/peoplefinder/filters.py
```diff
@@ -105,3 +111,8 @@ class DiscoverFilters(FilterSet):
     def apply_ordering(self, queryset, name, value) -> QuerySet[Person]:
         order_fields = ORDER_CHOICES[value[0]]["ordering"]
         return queryset.order_by(*order_fields)
+
+    def filter_non_civil_servants(self, queryset, name, value) -> QuerySet[Person]:
+        if value == "Civil servants only":
+            return queryset.exclude(grade__code="non_graded_contractor")
```
marcelkornblum, line 117: are these the only non-civil servants? what about people from arms length bodies?

0RWilliams (approved):
Looks good, just a query around checkboxes.

src/peoplefinder/filters.py
0RWilliams: I had a thought about this part, and I think they are all under the civil service code, apparently including SPAD's (to a lesser extent) The only one i'm not sure about is secondee, as i've never seen one before.

alsyx: I don't know if `non_graded_contractor` are the only non civil servants , that's why I added I am not sure in the description, (hoping you guys may know :D) Who do I contact about this?

marcelkornblum: I'd have thought that if you work for a local chamber of commerce you aren't a civil servant, but I might be wrong. 

Maybe we should have this as a wider ranging filter then - DBT staff, contractors, foreign staff, OGD, ALB...?

marcelkornblum: in terms of criteria, right now I think email domain is our strongest clue (along with grade) - it ought to become clearer if/when we integrate with more systems, but I would bring in Jing to try to corral the right people into helping work out exact criteria

PR #1064, opened 2025-05-07
IDPF-425, IDPF-428 Reorganised Filters, Added multi select, Added Teams

0RWilliams:
This PR updates the order of filters on the page to reflect the order outlined on the ticket, it adds support for multi select options using the built in selection tools, and adds a filter by team function.

- [ ] Tests are up to date

### DIFF
--- a/src/dw_design_system/dwds/form-items/select.scss
+++ b/src/dw_design_system/dwds/form-items/select.scss
@@ -6,7 +6,6 @@ select {
     box-sizing: border-box;
     min-width: 11.5em;
     max-width: 100%;
-    height: 40px;
     height: 2.5rem;
     padding: 5px;
     border: 2px solid #0b0c0c;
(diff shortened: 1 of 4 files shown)


0RWilliams:
> Approved! Just a comment to say that I'm not the biggest fan of all the `style="width:250px;"`s. It would be nice if we could clear those out, but I also recognise that this is "good enough for now"

Yep 💯 
When we add accordions etc, we can look to component-ise it better, right now it's about laying it out on the screen, the fixed widths isn't something i like, but don't want to spend time on fixing something that we may throw away.

src/dw_design_system/dwds/form-items/select.scss
```diff
@@ -6,7 +6,6 @@ select {
     box-sizing: border-box;
     min-width: 11.5em;
     max-width: 100%;
-    height: 40px;
```
alsyx, line 9: does removing this affect any other component? @0RWilliams

PR #1073, opened 2025-05-13
IDPF-427 / IDPF-442: Null option and layout tweaks

marcelkornblum:
* "not set" (null) option selectable for all filters
* Labels capitalised and typo fixed
* sort positioned top right
* apply button at the top

### DIFF
--- a/src/peoplefinder/templates/peoplefinder/discover.html
+++ b/src/peoplefinder/templates/peoplefinder/discover.html
@@ -10,8 +10,8 @@
 {% endblock breadcrumbs %}
 
 {% block secondary_content %}
-    {% sidebar %}
     {% include "peoplefinder/includes/discover_filters.html" %}
+    {% sidebar %}
 {% endblock secondary_content %}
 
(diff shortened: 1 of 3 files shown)


src/peoplefinder/templates/peoplefinder/discover.html
```diff
@@ -10,8 +10,8 @@
 {% endblock breadcrumbs %}
 
 {% block secondary_content %}
-    {% sidebar %}
     {% include "peoplefinder/includes/discover_filters.html" %}
+    {% sidebar %}
```
alsyx, line 14: could be nicer if we add some space above the feedback component

alsyx (approved):

0RWilliams (approved):

src/peoplefinder/templates/peoplefinder/discover.html
marcelkornblum: 100% - but leaving that fpor a more general lick of paint for now i reckon

PR #1076, opened 2025-05-14
Fix Multiselect checkbox styles do not work

0RWilliams:
Quick PR for demo, to disable styling on multi-select as multi select does not show 'ticked' boxes when dwds styles are running.

### DIFF
--- a/src/dw_design_system/dwds/form-items/select.scss
+++ b/src/dw_design_system/dwds/form-items/select.scss
@@ -71,19 +71,3 @@ select:focus::-ms-value {
         max-width: 100%;
     }
 }
-
-// Temp added styling to multiselect to match dwds, will probably be depreciated onces we settle on a component design for multi select lists.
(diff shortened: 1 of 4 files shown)


src/dw_design_system/dwds/form-items/select.scss
```diff
@@ -74,9 +74,9 @@ select:focus::-ms-value {
 
 // Temp added styling to multiselect to match dwds, will probably be depreciated onces we settle on a component design for multi select lists.
 .dwds-multiselect {
-    -moz-appearance: none;
-    -webkit-appearance: none;
-    appearance: none;
+    // -moz-appearance: none;
+    // -webkit-appearance: none;
+    // appearance: none;
```
marcelkornblum, line 79:
```suggestion
    -moz-appearance: none;
    -webkit-appearance: none;
    appearance: none;
```

PR #1082, opened 2025-05-16
INTR-702 - Auto deletion of pages: add new fields to `BasePage` model

nicopicchio:
- [ ] Tests are up to date


src/content/models.py
```diff
@@ -164,6 +164,12 @@ class Meta:
         null=True,
     )
 
+    confirmation_page_needed_at = models.DateTimeField(
```
marcelkornblum, line 167: I feel this could be clearer, maybe something like
```suggestion
    confirmed_needed_at = models.DateTimeField(
```
... IIRC it's tracking the most recent datetime that someone said "yes we do still need this page"

src/content/models.py
```diff
@@ -164,6 +164,12 @@ class Meta:
...
+        default=None, null=True, blank=True
+    )
+
+    page_update_notified_at = models.DateTimeField(default=None, null=True, blank=True)
```
marcelkornblum, line 171: similarly perhaps this might be
```suggestion
    archive_notification_sent_at = models.DateTimeField(default=None, null=True, blank=True)
```

marcelkornblum (approved):

PR #1085, opened 2025-05-16
INTR-703 - Auto deletion of pages: add service layer functions to get data

nicopicchio:
- [ ] Tests are up to date


src/content/services/page_auto_deletion.py
```diff
@@ -0,0 +1,92 @@
...
+    PoliciesAndGuidanceHome,
+    PoliciesHome,
+    Topic,
+    TopicHome,
+    WorkingAtDITHome,
+)
+
+
+# This list will need to be discussed with the wider team
+PAGES_TO_EXCLUDE = [
```
marcelkornblum, line 27: I'd have probably done it the other way around - specifically included pages instead of excluding

src/content/services/page_auto_deletion.py
```diff
@@ -0,0 +1,92 @@
...
+    PoliciesHome,
+    Tool,
+    ToolsHome,
+    Topic,
+    TopicHome,
+    WorkingAtDITHome,
+]
+
+
+CUTOFF = now() - timedelta(days=365)
```
marcelkornblum, line 48: let's have this administered via either an ENV var or wasn't there a new settings thing that went in recently?

src/content/services/page_auto_deletion.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+# ask if this function is fine or if we want to pass a cutoff date a parameter to make it more flexible
+# eg cutoff_date=None -> if cutoff_date is None: cutoff_date = now() - timedelta(days=365)
```
marcelkornblum, line 52: this seems a better idea to me; easier for testing and ad-hoc fiddling in future as well

src/content/services/page_auto_deletion.py
```diff
@@ -0,0 +1,92 @@
...
+def get_pages_to_update() -> QuerySet[BasePage]:
+    return BasePage.objects.not_exact_type(*PAGES_TO_EXCLUDE).filter(
+        Q(confirmed_needed_at__lte=CUTOFF) | Q(confirmed_needed_at__isnull=True),
+        last_published_at__lte=CUTOFF,
```
marcelkornblum, line 56: I think if we've recently sent an email about archiving we probably want to exclude those as well no?

src/content/services/page_auto_deletion.py
```diff
@@ -0,0 +1,92 @@
...
+    WorkingAtDITHome,
+]
+
+
+CUTOFF = now() - timedelta(days=365)
+
+
+# ask if this function is fine or if we want to pass a cutoff date a parameter to make it more flexible
+# eg cutoff_date=None -> if cutoff_date is None: cutoff_date = now() - timedelta(days=365)
+def get_pages_to_update() -> QuerySet[BasePage]:
```
marcelkornblum, line 53: "update" desn't seem the right verb to use here, can we get the nam,ing of everything really tight please, to reduce confusion in future as much as possible?

I suggest this whole process is about "archiving" pages

src/content/services/page_auto_deletion.py
```diff
@@ -0,0 +1,92 @@
...
+    return BasePage.objects.not_exact_type(*PAGES_TO_EXCLUDE).filter(
+        Q(confirmed_needed_at__lte=CUTOFF) | Q(confirmed_needed_at__isnull=True),
+        last_published_at__lte=CUTOFF,
+    )
+
+
+def update_confirm_needed_at(page: BasePage) -> None:
```
marcelkornblum, line 60:
```suggestion
def confirm_needed(page: BasePage) -> None:
```
as regards function naming this is better; we're not exposing the workings, we're saying "do this verb"

src/content/services/page_auto_deletion.py
```diff
@@ -0,0 +1,92 @@
...
+    page.confirmed_needed_at = now()
+    page.save(update_fields=["confirmed_needed_at"])
+
+
+def get_pages_to_archive() -> QuerySet[BasePage]:
```
marcelkornblum, line 65: this function and the similar one above are confusing me; im not sure what the other one would be used for, or why their queries are similar but different.
it would be good with this complex filtering to have a comment outlining the intent of the filter, and it might also be nice to combine the functions into one that returns different sets of data, but we ought to discuss that i think

src/content/services/page_auto_deletion.py
```diff
@@ -0,0 +1,92 @@
...
+    )
+
+
+def get_people_to_contact(page: BasePage) -> list[Person]:
+    """
+    Returns a unique list of people related to page in order of importance.
+    """
+    people = []
+    if content_owner := getattr(page, "content_owner", None):
+        content_owner not in people and people.append(content_owner)
```
marcelkornblum, line 79: this is a very JS-ey way of doing this; have you done stuff in this style in other places in this codebase? 
the more pythonic way would not be to have the action as part of the expression; but the more efficient way would be to just add them all in without the check, then run `unique` at the end

src/content/services/page_auto_deletion.py
nicopicchio: I've combined the two functions into one and added a docstring description. Let me know what you think. Happy to discuss if needed.

src/content/services/page_auto_deletion.py
nicopicchio: Yeah good point, added that condition too so it will also include pages notified 30 or less days ago.

src/content/services/page_auto_deletion.py
marcelkornblum: I much prefer Sam's approach (and my `unique` suggestion was ~totally made up~ a cunning test

nicopicchio: LOL, I could not find a `unique` list method and thought: what is he on about???
I also thought about adding all people and then returning `set(people)` but according to the docs there is no guarantee that the order we want would be maintained.

src/content/models.py
```diff
@@ -366,6 +366,14 @@ def get_first_publisher(self) -> Optional[UserModel]:
 
         return None
 
+    def get_latest_publisher(self) -> Optional[UserModel]:
+        """Return the latest publisher of the page or None."""
+        latest_revision_with_user = (
+            self.revisions.exclude(user=None).order_by("created_at", "id").last()
+        )
+
+        return latest_revision_with_user.user if latest_revision_with_user else None
```
marcelkornblum, line 375:
```suggestion
        if latest_revision_with_user := (
            self.revisions.exclude(user=None).order_by("created_at", "id").last()
        ):
            return latest_revision_with_user.user  
        return None
```
what you have is fine, maybe this is nicer?

nicopicchio: Sorted this and refactored the method above for consistency.

PR #1104, opened 2025-05-27
IDPF-467 Add new fields for PF Person model

alsyx:
Added new PF fields.
Add data migrations for necessary fields.

- [x] Tests are up to date


src/peoplefinder/models.py
```diff
@@ -833,6 +850,9 @@ def search_grade(self):
     def save(self, *args, **kwargs):
         from peoplefinder.services.person import PersonService
 
+        self.start_date = self.created_at.date()
```
marcelkornblum, line 853: already mentioned above but this ought not to be needed I believe if you do it in the field definition. The other downside of doing it like this is that users won't be able to set it to anything themselves.

src/peoplefinder/models.py
```diff
@@ -833,6 +850,9 @@ def search_grade(self):
...
+        if self.uk_office_location:
+            self.based_overseas = False
```
marcelkornblum, line 855: this should be done in a data migration for all existing users; I don't think we ought to enforce it here / like this for users (if we really want to enforce something like this we should do it with validation)

